PR #133, opened 2023-11-22
Primitive statics support fixes

CaelmBleidd:


usvm-jvm/src/main/kotlin/org/usvm/api/util/JcTestStateResolver.kt
```diff
@@ -286,25 +335,30 @@ abstract class JcTestStateResolver<T>(
...
-        if (stringConstant != null) {
-            return decoderApi.createStringConst(stringConstant.key)
-        }
-
         val valueField = ctx.stringValueField
         val strValueLValue = UFieldLValue(ctx.typeToSort(valueField.fieldType), ref, valueField.field)
-        val strValue = resolveLValue(strValueLValue, valueField.fieldType)
+
+        val strValue = if (isStaticHeapRef(ref)) {
+            withMode(ResolveMode.CURRENT) {
```
Saloed, line 355: Why we use `withMode` here instaed of `finalStateMemory`?

usvm-jvm/src/main/kotlin/org/usvm/api/util/JcTestStateResolver.kt
```diff
@@ -150,7 +199,7 @@ abstract class JcTestStateResolver<T>(
             model.typeStreamOf(ref)
         } else {
             // allocated object
-            memory.typeStreamOf(ref)
+            finalStateMemory.typeStreamOf(ref)
```
Saloed, line 202: Why we use `finalStateMemory` here? Maybe it is better to read allocated ref from memory and receive an exception in the case of the model?

Saloed (changes requested):

usvm-jvm/src/main/kotlin/org/usvm/api/util/JcTestStateResolver.kt
CaelmBleidd: Because sometimes we resolve strings in the model. If we read here from the final memory, only the address will be correct, but the content of the string will be constructed from the model. That led us to an inconsistent state

usvm-jvm/src/main/kotlin/org/usvm/machine/interpreter/statics/JcStaticFieldsRegion.kt
```diff
@@ -0,0 +1,148 @@
...
+
+        val mutablePrimitiveStaticFieldsToSymbolicValues = staticFields
+            .entries
+            .filter { (field, _) -> fieldShouldBeSymbolic(field) }
+            .associate { (field, value) ->
+                val lvalue = JcStaticFieldLValue(field, value.sort)
+                val regionId = lvalue.memoryRegionId as JcStaticFieldRegionId
+                val symbol = state.ctx.mkStaticFieldReading(regionId, field, value.sort)
+
+                state.memory.write(lvalue, symbol, guard = state.ctx.trueExpr)
```
Saloed, line 89: Why do we need to write this symbol to the memory? Also, it is seems like we can simply remove all such fields from the region, instead of writing a bunch of symbols

CaelmBleidd: Fixed

Saloed (approved):

PR #166, opened 2023-12-15
Allow configuring how jar paths are retrieved

IlyaMuravjov:
This PR allows changing of how paths to jars and Java are retrieved. Previously path were always retrieved by reading environment variables, but in UtBot we need to dynamically change these paths without starting a new process.

### DIFF
--- a/usvm-jvm-instrumentation/src/main/kotlin/org/usvm/instrumentation/util/Constants.kt
+++ b/usvm-jvm-instrumentation/src/main/kotlin/org/usvm/instrumentation/util/Constants.kt
@@ -23,14 +23,6 @@ object InstrumentationModuleConstants {
 
     const val nameForExistingButNullString = "USVM_GENERATED_NULL_STRING"
 
-    //Passes as environment parameter
-    val pathToUsvmInstrumentationJar: String
-        get() = System.getenv("usvm-jvm-instrumentation-jar")
-
-    val pathToUsvmCollectorsJar: String
-        get() = System.getenv("usvm-jvm-collectors-jar")
-
-    val pathToJava: String
-        get() = System.getenv()["JAVA_HOME"] ?: System.getProperty("java.home")
-
+    //Environment variable used to pass path to collectors jar from main process to instrumented process
+    val envVarForPathToUsvmCollectorsJarPath = "usvm-jvm-collectors-jar"
 }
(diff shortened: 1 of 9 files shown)


usvm-jvm-instrumentation/src/main/kotlin/org/usvm/instrumentation/util/Constants.kt
```diff
@@ -24,13 +24,14 @@ object InstrumentationModuleConstants {
     const val nameForExistingButNullString = "USVM_GENERATED_NULL_STRING"
 
     //Passes as environment parameter
-    val pathToUsvmInstrumentationJar: String
-        get() = System.getenv("usvm-jvm-instrumentation-jar")
+    var pathToUsvmInstrumentationJarGetter = { System.getenv("usvm-jvm-instrumentation-jar") }
```
Saloed, line 27: Can you explain please how such getters allow you to change various `path`s of a running process?

IlyaMuravjov: When currently running main process needs to start a **new** instrumented process, it will access relevant `path`s via these getters.

To be more clear:
* these getters will be set by UtBot before any other call to USVM API,
* these getters will be set by UtBot to functions that always return the same value.
 
The only reason getters are used it to allow for lazy jar extraction that is only triggered during first getter call. Although, if an additional layer of indirection makes it look like these getters can somehow change `path`s that are used by already started instrumented process (which they can't), then I think it's better to just have mutable paths here, since possibly unnecessary eager jar extraction is not all that expensive.

Motivation for this change is that previously specifying these `path`s required either:
* knowing them before main process is started,
* changing environment variables of currently running main process in some obscure way via reflection.

Saloed: Maybe it is better to remove all these environment variables and provide all required paths via `InstrumentationProcessRunner` constructor?

IlyaMuravjov: I agree, that it's better to inject settings via constructors and not via global state. However, there's a `// TODO move in common settings file` on `object InstrumentationModuleConstants` implying that some refactoring is probably going to take place there. So I minimized my changes to avoid conflicts with that refactoring.

Saloed: No refactoring is in progress, so feel free to make any changes

Saloed (approved):

PR #175, opened 2024-02-13
Improve USVM performance

Saloed:
1. Use non-linked persistent hash maps. 
2. Use persistent hash map builders to improve bulk mutations.
3. Introduce `UModelEvaluator` to allow model completion strategy customization. Such strategies can significantly improve performance in some usage scenarios. For example, we can complete addresses with concrete refs instead of null.
4. Few minor fixes.


CaelmBleidd (approved):
LGTM

PR #176, opened 2024-02-14
Log failed asserts and forks

Saloed:


usvm-core/src/main/kotlin/org/usvm/utils/StateUtils.kt
```diff
@@ -97,3 +99,11 @@ fun <T : UState<Type, *, *, *, *, T>, Type> T.applySoftConstraints() {
         }
     }
 }
+
+inline fun StepScope<*, *, *, *>.onStateDeath(crossinline body: () -> Unit) {
+    if (isDead) body()
+}
+
+inline fun Unit?.logAssertFailure(crossinline message: () -> String) = apply {
```
CaelmBleidd, line 107: Please, specify the returning type

usvm-core/src/main/kotlin/org/usvm/api/collection/ObjectMapCollectionApi.kt
```diff
@@ -61,7 +62,9 @@ object ObjectMapCollectionApi {
                 with(ctx) {
                     val boundConstraint = mkSizeGeExpr(length, mkSizeExpr(0))
                     // Map size must be correct regardless of guard
-                    assert(boundConstraint) ?: return null
+                    assert(boundConstraint)
+                        .logAssertFailure { "SymbolicMap size correctness constraint" }
```
CaelmBleidd, line 66: Let's add a common word for all these messages to be able to find them in logs files. Something like "Constaint violation: Symbolic..."

CaelmBleidd (approved):
LGTM

PR #178, opened 2024-02-19
Track state fork points

Saloed:


usvm-core/src/main/kotlin/org/usvm/StepScope.kt
```diff
@@ -132,18 +145,32 @@ class StepScope<T : UState<Type, *, Statement, Context, *, T>, Type, Statement,
 
         // Interpret the first state as original and others as forked
         this.forkedStates += forkedStates.subList(1, forkedStates.size)
+
+        if (forkedStates.size > 1 || registerForkPoint) {
+            forkedStates.forEach { it.forkPoints += possibleForkPoint }
```
CaelmBleidd, line 150: Is it expected that `registerForkPoint = false` doesn't affect this line in case of the fork? If so, this should be highlighted in the documentation

usvm-core/src/main/kotlin/org/usvm/StepScope.kt
```diff
@@ -180,12 +207,15 @@ class StepScope<T : UState<Type, *, Statement, Context, *, T>, Type, Statement,
             return fork(condition, blockOnTrueState, blockOnFalseState)
         }
 
+        // If condition is concrete there is no fork point possibility
+        val registerForkPoint = !condition.isConcrete
```
CaelmBleidd, line 211: Why there is no option to disable fork registration at all as it made in other methods?

CaelmBleidd (changes requested):

usvm-core/src/main/kotlin/org/usvm/StepScope.kt
Saloed: I renamed `registerForkPoint` to `alwaysRegisterForkPoint`. The logic behind fork points regitstration should be more clear now.

usvm-core/src/main/kotlin/org/usvm/StepScope.kt
Saloed: There is no such option for other methods. This should be more clear after renaming

CaelmBleidd (approved):

PR #182, opened 2024-02-21
JcMachine debug profiler

Saloed:


usvm-jvm/src/main/kotlin/org/usvm/machine/JcDebugProfileObserver.kt
```diff
@@ -0,0 +1,187 @@
...
+
+    private fun StringBuilder.reportMethodStats() {
+        val methodStats = aggregateStats()
+
+        appendLine("Method stats:")
+        for ((cls, methods) in methodStats) {
+            appendLine("-".repeat(20))
+            appendLine("$cls")
+            for ((method, cnt) in methods) {
+                appendLine("$method | $cnt | ${methodCalls[method]?.values?.sum()}")
```
CaelmBleidd, line 151: Is it possible that we don't have the method in the map? If not, probably it is better to use `getValue`, otherwise, i don't think `null` is a good thing to show here

CaelmBleidd (changes requested):

CaelmBleidd (approved):

PR #191, opened 2024-07-05
Remove dependency on 'jvm-dataflow' in 'usvm-jvm' module

Lipen:
This PR inlines `JcApplicationGraphImpl` (from `usvm-dataflow-jvm` module, originally from `jacodb-analysis`) into `JcApplicationGraph` (in `uvsm-jvm` module, package `org.usvm.machine`). This allows for removing the dependency on `usvm-dataflow-jvm` module in `usvm-jvm`.

### DIFF
--- a/usvm-jvm/build.gradle.kts
+++ b/usvm-jvm/build.gradle.kts
@@ -26,7 +26,6 @@ val approximationsVersion = "0f081f101e"
 
 dependencies {
     implementation(project(":usvm-core"))
-    implementation(project(":usvm-jvm-dataflow"))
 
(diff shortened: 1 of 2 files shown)


sergeypospelov (approved):
Why is it bad to have an explicit dependency on `usvm-jvm-dataflow`? If you change the implementation of `JcApplicationGraphImpl` to efficient one, you'll have to copy-paste it, won't you?

Lipen:
Figured out in personal commutations that `usvm-*` in fact _should_ depend on `usvm-dataflow-*`, as USVM provides (or rather aims to provide) not only the SE-core, but also dataflow-based solutions. Currently, only `JcApplicationGraph` is used from here, but there will be more functionality reused as the development goes on.

PR #197, opened 2024-07-18
Introduction of TypeScript symbolic machine 

CaelmBleidd:
An initial commit for TS symbolic virtual machine, the first implementation of the main interfaces

### DIFF
--- a/usvm-ts/src/test/kotlin/org/usvm/util/TSMethodTestRunner.kt
+++ b/usvm-ts/src/test/kotlin/org/usvm/util/TSMethodTestRunner.kt
@@ -126,7 +126,7 @@ open class TSMethodTestRunner : TestRunner<TSTest, MethodDescriptor, EtsType?, T
         get() = TODO("Not yet implemented")
 
     private fun getProject(fileName: String): EtsFile {
-        val jsonWithoutExtension = "/ir/${fileName}.json"
+        val jsonWithoutExtension = "/ir/$fileName.json"
         val sampleFilePath = javaClass.getResourceAsStream(jsonWithoutExtension)
             ?: error("Resource not found: $jsonWithoutExtension")
 
@@ -141,7 +141,6 @@ open class TSMethodTestRunner : TestRunner<TSTest, MethodDescriptor, EtsType?, T
 
         return cls.methods.find { it.name == desc.methodName && it.parameters.size == desc.argumentsNumber }
             ?: error("No method matching $desc found in $name")
-
     }
 
     override val runner: (MethodDescriptor, UMachineOptions) -> List<TSTest>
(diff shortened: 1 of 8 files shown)


usvm-ts/src/main/kotlin/org/usvm/TSApplicationGraph.kt
```diff
@@ -0,0 +1,36 @@
...
+        applicationGraph.entryPoints(method)
+
+    override fun exitPoints(method: EtsMethod): Sequence<EtsStmt> =
+        applicationGraph.exitPoints(method)
+
+    override fun methodOf(node: EtsStmt): EtsMethod =
+        applicationGraph.methodOf(node)
+
+    override fun statementsOf(method: EtsMethod): Sequence<EtsStmt> = sequence {
+        method.cfg.stmts.forEach { yield(it) }
```
Saloed, line 34: Can we avoid yield here?

usvm-ts/src/test/kotlin/org/usvm/util/TSMethodTestRunner.kt
```diff
@@ -0,0 +1,178 @@
...
+import org.usvm.test.util.TestRunner
+import org.usvm.test.util.checkers.ignoreNumberOfAnalysisResults
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.milliseconds
+
+open class TSMethodTestRunner : TestRunner<TSTest, MethodDescriptor, EtsType?, TSMethodCoverage>() {
+
+    protected val globalClassName = "_DEFAULT_ARK_CLASS"
+
+    protected inline fun <reified R> discoverProperties(
```
Saloed, line 22: Do we plan to somehow translate TS objects into Java objects?

usvm-ts/src/test/kotlin/org/usvm/util/TSMethodTestRunner.kt
```diff
@@ -0,0 +1,178 @@
...
+                typeTransformer(T4::class),
+                typeTransformer(R::class)
+            ),
+            checkMode = CheckMode.MATCH_PROPERTIES,
+            coverageChecker = { _ -> true }
+        )
+    }
+
+    override val typeTransformer: (Any?) -> EtsType
+        get() = TODO("Not yet implemented")
```
Saloed, line 123: Is it ok?

Saloed (changes requested):

usvm-ts/src/test/kotlin/org/usvm/util/TSMethodTestRunner.kt
CaelmBleidd: Yes, in the future they will be represented probably as a map from a property name to its value

usvm-ts/src/test/kotlin/org/usvm/util/TSMethodTestRunner.kt
CaelmBleidd: Yes, all tests for TS are disabled now and require additional work to enable. It will be done in further requests

Saloed (approved):

PR #200, opened 2024-07-29
TS symbolic machine expansion

zishkaz:
* TSExprResolver skeleton with TSSimpleValueResolver
* TSTypeSystem
* TSInterpreter skeleton


usvm-ts/src/main/kotlin/org/usvm/TSTypeSystem.kt
CaelmBleidd, line 1: These are too complicated changes to be made like this. I expect to have a separate PR with lot of tests for this

CaelmBleidd (changes requested):
There are no tests in the request, you need to provide them

PR #201, opened 2024-07-29
TS symbolic machine expansion

zishkaz:
* TSExprResolver skeleton with TSSimpleValueResolver
* TSInterpreter skeleton
* Other miscellaneous implementations


usvm-ts/src/test/kotlin/org/usvm/util/TSMethodTestRunner.kt
```diff
@@ -91,9 +91,10 @@ open class TSMethodTestRunner : TestRunner<TSTest, MethodDescriptor, EtsType?, T
 
     override val typeTransformer: (Any?) -> EtsType
         get() = {
-            require(it is KClass<*>) { "Only TSObjects are allowed" }
+            // Both KClass and TSObject instances come here
+            val temp = if (it is KClass<*> || it == null) it else it::class
```
CaelmBleidd, line 95: How is it possible? Isn't it a mistake? How a TSObject could come here if the signature for it is

```
  protected inline fun <reified T : TSObject, reified R : TSObject> 
  ...
  arrayOf(typeTransformer(T::class), typeTransformer(R::class))
```
?

usvm-ts/src/test/kotlin/org/usvm/util/TSMethodTestRunner.kt
```diff
@@ -107,8 +108,12 @@ open class TSMethodTestRunner : TestRunner<TSTest, MethodDescriptor, EtsType?, T
             }
         }
 
+    /*
+        For now type checks are disabled for development purposes
+        TODO: implement
```
CaelmBleidd, line 113: create an issue for it and link it with this place

usvm-ts/src/test/kotlin/org/usvm/util/TSTestResolver.kt
```diff
@@ -1,12 +1,90 @@
...
-@Suppress("UNUSED_PARAMETER")
 class TSTestResolver {
-    fun resolve(method: EtsMethod, state: TSState): TSTest {
-        TODO()
+
+    fun resolve(method: EtsMethod, state: TSState): TSTest = with(state.ctx) {
+        val model = state.models.first()
+        when (val methodResult = state.methodResult) {
+            is TSMethodResult.Success -> {
+                val returnValue = resolveExpr(model.eval(methodResult.value), method.returnType)
```
CaelmBleidd, line 32: lets avoid nested calculations in complicated places to make it easier for debug. For example, here it is easier if the expr is evaluated at the previous instruction.

```kotlin
val valueToResolve = model.eval(methodResult.value)
val returnValue = resolveExpr(valueToResolve, method.returnType)
```

usvm-ts/src/test/kotlin/org/usvm/samples/Arguments.kt
```diff
@@ -19,4 +18,41 @@ class Arguments : TSMethodTestRunner() {
             )
         )
     }
+
+    @Test
+    fun testSingleArg() {
+        discoverProperties<TSObject.TSNumber>(
```
CaelmBleidd, line 24: Please, add properties matchers to tests. Otherwise, for now they don't check anything but the machine produced something

CaelmBleidd (changes requested):

usvm-ts/src/test/kotlin/org/usvm/util/TSMethodTestRunner.kt
zishkaz: typeTransformer is used in parent class _TestRunner_ in method _checkType_:
<pre>
private fun checkTypes(
        argumentsTypes: Array&lt;out Type&gt;,
        valuesToCheck: List&lt;List&lt;Any?&gt;&gt;,
    ) {
        for (valueToCheck in valuesToCheck) {
            val actualArgsTypes = valueToCheck.map { <b>typeTransformer</b>(it) }
            val expectedTypesWithActual = argumentsTypes.zip(actualArgsTypes)

           ...
        }
    }
</pre>

Which results in TSObject entity coming in TSMethodTestRunner _typeTransfomer_. Running any sample without this check will result in this error:

```java.lang.IllegalStateException: Should not be called: class org.usvm.TSObject$TSNumber$Double```

File: usvm-ts/src/main/kotlin/org/usvm/TSInterpreter.kt [hunk 5]
```diff
@@ -40,14 +57,117 @@ class TSInterpreter(
...
+        )
+
+    // (method, localName) -> idx
+    private val localVarToIdx = mutableMapOf<EtsMethod, MutableMap<String, Int>>()
+
+    private fun mapLocalToIdxMapper(method: EtsMethod, local: EtsValue) =
+        when (local) {
+            is EtsLocal -> {
+//                // See https://github.com/UnitTestBot/usvm/issues/202
+//                if (local.name == "this") 0 else
```
CaelmBleidd, line 157: Is it fixed or is it not fixed? Also, we should not create a commented code in a regular scenarios. Moreover, create comments not with "see", but with "TODO", because there is no fast way to check this sees

CaelmBleidd (changes requested):

CaelmBleidd (approved):

PR #204, opened 2024-08-07
Updated usvm-api and approximations

MchKosticyn:
- added encoder api
- added implementations for 'Engine' methods
- updated approximations

### DIFF
--- a/usvm-jvm/src/main/kotlin/org/usvm/machine/JcApproximations.kt
+++ b/usvm-jvm/src/main/kotlin/org/usvm/machine/JcApproximations.kt
@@ -70,6 +70,7 @@ import kotlin.reflect.KFunction0
 import kotlin.reflect.KFunction1
 import kotlin.reflect.KFunction2
 import kotlin.reflect.jvm.javaMethod
+import org.usvm.api.makeNullableSymbolicRefWithSameType
 
 class JcMethodApproximationResolver(
     private val ctx: JcContext,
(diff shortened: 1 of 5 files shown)


usvm-jvm/src/main/kotlin/org/usvm/machine/JcApproximations.kt
```diff
@@ -2,34 +2,8 @@ package org.usvm.machine
...
-import org.usvm.UBvSort
-import org.usvm.UConcreteHeapRef
-import org.usvm.UExpr
-import org.usvm.UFpSort
-import org.usvm.UHeapRef
-import org.usvm.api.Engine
-import org.usvm.api.SymbolicIdentityMap
-import org.usvm.api.SymbolicList
-import org.usvm.api.SymbolicMap
+import org.jacodb.api.jvm.*
```
Saloed, line 5: It is better to avoid such changes

usvm-jvm/src/main/kotlin/org/usvm/machine/JcApproximations.kt
```diff
@@ -110,66 +85,84 @@ class JcMethodApproximationResolver(
...
+        if (className == "jdk.internal.misc.Unsafe") {
             if (approximateUnsafeVirtualMethod(methodCall)) return true
         }
 
-        if (method.name == "clone" && method.enclosingClass == ctx.cp.objectClass) {
+        if (method.name == "clone" && enclosingClass == ctx.cp.objectClass) {
             if (approximateArrayClone(methodCall)) return true
         }
 
+        if (className == "org.apache.commons.logging.Log") {
```
Saloed, line 122: Maybe we can remove non-standard classes from JcApproximations and write this part in `java-stdlib-approximations`?

usvm-jvm/src/main/kotlin/org/usvm/machine/JcApproximations.kt
```diff
@@ -110,66 +85,84 @@ class JcMethodApproximationResolver(
...
             return true
         }
 
-        if (method.enclosingClass == ctx.classType.jcClass) {
+        if (enclosingClass == ctx.classType.jcClass) {
             if (approximateClassStaticMethod(methodCall)) return true
         }
 
-        if (method.enclosingClass.name == "java.lang.System") {
+        if (className == "generated.java.lang.PrimitiveTypeUtils") {
```
Saloed, line 141: Approximation for the approximation?)

Saloed (changes requested):

usvm-jvm/src/main/kotlin/org/usvm/machine/JcApproximations.kt
MchKosticyn: Да, Class.getPrimitiveClass является приватным методом, поэтому в аппроксимациях примитивных классов используется PrimitiveTypeUtils

Saloed: Maybe we can write an approximation for `Class.getPrimitiveClass`?  I don't like that we have dependency on approximations implementation in USVM. So, i want to avoid `generated.java.lang.PrimitiveTypeUtils` here.

MchKosticyn: We already have approximation for Class.getPrimitiveClass

Saloed (approved):

PR #208, opened 2024-08-13
Updated approximations version

MchKosticyn:


usvm-jvm/src/test/kotlin/org/usvm/samples/JavaMethodTestRunner.kt
```diff
@@ -807,7 +807,7 @@ open class JavaMethodTestRunner : TestRunner<JcTest, KFunction<*>, KClass<*>?, J
         coverageZone = CoverageZone.TRANSITIVE,
         exceptionsPropagation = true,
         timeout = 60_000.milliseconds,
-        stepsFromLastCovered = 3500L,
+        stepsFromLastCovered = 4500L,
```
Saloed, line 810: Let's avoid global changes. We can override this property for a single test

Saloed (changes requested):

Saloed (approved):

PR #209, opened 2024-08-16
[python] Introducing virtual objects with multiple sets of slots

jefremof:
Allows to fix the problem with `PyMockTypeStream` when it doesn't start with `MockType`.
Now it is possible to create multiple virtual object types, the slots for which are specified using a mask.


usvm-python/usvm-python-main/src/main/java/org/usvm/interpreter/CPythonAdapter.java
```diff
@@ -76,7 +76,8 @@ public class CPythonAdapter {
     public native String getNameOfPythonType(long type);
     public static native int getInstructionFromFrame(long frameRef);
     public static native long getCodeFromFrame(long frameRef);
-    public native long allocateVirtualObject(VirtualPythonObject object);
+    public native long allocateRawVirtualObject(VirtualPythonObject object);
+    public native long allocateVirtualObject(VirtualPythonObject object, byte[] mask);
```
tochilinak, line 80: Could you please add some kind of mapping from type slot to a bit number that corresponds to it?

For example, you can use an approach similar to [`SymbolicMethodId`](https://github.com/UnitTestBot/usvm/blob/main/usvm-python/usvm-python-annotations/src/main/kotlin/org/usvm/annotations/ids/SymbolicMethodId.kt):

1. Introduce enum values for type slots. They should contain the name of the slot id (like `Py_nb_add`). This value will be used in generated C code. Also, add a field for required bit number. This value should be set from C code.
2. Generate code that sets the bit numbers. [Here is the code](https://github.com/UnitTestBot/usvm/blob/main/usvm-python/usvm-python-annotations/src/main/kotlin/org/usvm/annotations/codegeneration/SymbolicMethodGeneration.kt#L37) where `cRef`s of `SymbolicMethodId` are set. That generated macro is inserted into generated header `SymolicMethods.h` [here](https://github.com/UnitTestBot/usvm/blob/37050f9502fdbbf2f4de66b447086c0bc5a6a641/usvm-python/usvm-python-annotations/src/main/kotlin/org/usvm/annotations/SymbolicMethodProcessor.kt#L31). You can just add your macro there as well.
3. Finally, the generated macro can be called during python initialization (like [here](https://github.com/UnitTestBot/usvm/blob/37050f9502fdbbf2f4de66b447086c0bc5a6a641/usvm-python/cpythonadapter/src/main/c/org_usvm_interpreter_CPythonAdapter.c#L84)).
4. Each [TypeProtocol](https://github.com/UnitTestBot/usvm/blob/37050f9502fdbbf2f4de66b447086c0bc5a6a641/usvm-python/usvm-python-main/src/main/kotlin/org/usvm/machine/types/VirtualTypes.kt#L43) should then refer to one of the type slot id's so that later this could be used to generate needed masks.

tochilinak:
Also, please add some unit tests for generated virtual objects. For example, add a test where you create an object without `nb_add`, and check with `ConcretePythonInterpreter` that its type doesn't have it indeed.

You might need [this function](https://github.com/UnitTestBot/usvm/blob/main/usvm-python/usvm-python-main/src/main/kotlin/org/usvm/machine/interpreters/concrete/ConcretePythonInterpreter.kt#L151)

usvm-python/cpythonadapter/src/main/c/virtual_objects.c
```diff
@@ -229,44 +236,151 @@ tp_setattro(PyObject *self, PyObject *attr, PyObject *value) {
...
-        Virtual_mp_ass_subscript,
-        {0, 0}
-    };
+initialize_virtual_object_ready_types() {
+    ready_virtual_object_types = PyDict_New();
+}
+
+void
+deinitialize_virtual_object_ready_types() {
+    Py_DECREF(ready_virtual_object_types);
```
tochilinak, line 262: Do not forget to set `ready_virtual_object_types` to `nullptr` after this.

usvm-python/cpythonadapter/src/main/c/virtual_objects.c
```diff
@@ -229,44 +236,151 @@ tp_setattro(PyObject *self, PyObject *attr, PyObject *value) {
...
+        // COUNT_OR_SET(sq_concat);
+        // just use the macro again to add a new slot
+    }
+    slots[i++] = final_slot;
+    AVAILABLE_SLOTS = slots;
+}
+
+void
+deinitialize_virtual_object_available_slots() {
+    PyMem_RawFree(AVAILABLE_SLOTS);
```
tochilinak, line 299: Do not forget to set `AVAILABLE_SLOTS` to `nullptr` after this.

usvm-python/cpythonadapter/src/main/c/virtual_objects.c
```diff
@@ -278,6 +392,25 @@ allocate_raw_virtual_object(JNIEnv *env, jobject object) {
     return (PyObject *) result;
 }
 
+PyObject *
+allocate_virtual_object(JNIEnv *env, jobject object, jbyteArray mask) {
```
tochilinak, line 396: The difference between "raw" virtual object and an ordinary one was that the latter had fields `ctx` and `adapter` initialized. The thing is, we can set those fields only after calling `concolicRun`, but virtual objects must be initialized before that. The object that you create here is also raw, meaning that we need to call `finish_virtual_object_initialization` on it later. So, please, return the word "raw" to avoid confusion. You may call the other function something like `allocate_raw_virtual_object_with_all_slots`.

tochilinak (changes requested):
Left the comments above.

usvm-python/usvm-python-main/src/main/kotlin/org/usvm/machine/types/VirtualTypes.kt
```diff
@@ -113,42 +114,42 @@ object HasMpLength : TypeProtocol() {
...
 }
 
-object HasTpCall : TypeProtocol() {
+object HasTpCall : TypeProtocol(SlotId.TpCall) {
     override fun acceptsConcrete(type: ConcretePythonType): Boolean =
         ConcretePythonInterpreter.typeHasTpCall(type.asObject)
 }
 
-object HasTpHash : TypeProtocol() {
+object HasTpHash : TypeProtocol(SlotId.TpHash) {
```
tochilinak, line 152: Please add this comment for mandatory slots:

```
TODO: since we cannot turn off this slot on virtual object, we may need to remove this [TypeProtocol] in the future.
```

usvm-python/usvm-python-annotations/src/main/kotlin/org/usvm/annotations/ids/SlotId.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    companion object {
+        init {
+            values().filter {!it.mandatory}.forEachIndexed {
+                index, entry -> entry.maskBit = index
+            }
+        }
+    }
+    private var maskBit: Int? = null
+    fun getMaskBit(): Int = maskBit!!
```
tochilinak, line 46: Instead of using `!!`, please use `?: error("Error message. Like 'cannot call this on mandatory slots'")`.

usvm-python/usvm-python-main/src/main/kotlin/org/usvm/machine/interpreters/concrete/ConcretePythonInterpreter.kt
```diff
@@ -162,7 +162,18 @@ object ConcretePythonInterpreter {
     }
 
     fun allocateVirtualObject(virtualObject: VirtualPythonObject): PyObject {
-        val ref = pythonAdapter.allocateVirtualObject(virtualObject)
+        /*
+         * Usage example:
```
tochilinak, line 166: I believe, now this `usage example` should be modified to use your new auxiluary methods.

tochilinak (changes requested):

usvm-python/usvm-python-main/src/main/kotlin/org/usvm/machine/types/VirtualTypes.kt
```diff
@@ -40,7 +41,7 @@ data class ConcreteTypeNegation(val concreteType: ConcretePythonType) : VirtualP
     }
 }
 
-sealed class TypeProtocol : VirtualPythonType() {
+sealed class TypeProtocol(private val slotId: SlotId? = null) : VirtualPythonType() {
```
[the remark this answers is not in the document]
jefremof: This is a temporary solution, since some slots, such as `nb_int`, `nb_index` and `mp_length`, are missing their implementation in `virtual_objects.c`.

usvm-python/usvm-python-annotations/src/main/kotlin/org/usvm/annotations/ids/SlotId.kt
```diff
@@ -0,0 +1,69 @@
...
+ * 
+ * The usage of swapSlotBit or setSlotBit with these slots
+ * will not have any effect on the mask.
+ */
+enum class SlotId(
+    val slotName: String,
+    val mandatory: Boolean = false
+) {
+    TpGetattro("tp_getattro"),
+    TpSetattro("tp_setattro", true),
```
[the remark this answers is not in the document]
jefremof: The tests in the `VirtualObjectsTest.kt` showed that `tp_getattro` can be disabled, but that's not the case with `tp_setattro` . So, yeah, I'm sure. It can be made mandatory, but I tend not to add constraints unnecessarily.

usvm-python/usvm-python-main/src/main/kotlin/org/usvm/machine/types/VirtualTypes.kt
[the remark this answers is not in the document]
tochilinak: Ok. Please, add a comment about that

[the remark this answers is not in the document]
jefremof: Already done it

usvm-python/usvm-python-main/src/main/kotlin/org/usvm/machine/interpreters/concrete/ConcretePythonInterpreter.kt
jefremof: Done

usvm-python/usvm-python-annotations/src/main/kotlin/org/usvm/annotations/ids/SlotId.kt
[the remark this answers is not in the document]
jefremof: No, there was a bug in the test. Fixed that, `tp_getattro` is now mandatory too.

tochilinak:
Fine with me. Waiting from review from @sergeypospelov.

usvm-python/usvm-python-main/src/main/kotlin/org/usvm/machine/interpreters/concrete/utils/VirtualPythonObject.kt
```diff
@@ -1,6 +1,10 @@
 package org.usvm.machine.interpreters.concrete.utils
 
+const val MAX_NEEDED_MASK_BYTE_NUMBER: Int = 12
+const val ALL_SLOTS_BYTE: Int = 0b11111111
```
sergeypospelov, line 4: Can we make it private?

usvm-python/usvm-python-main/src/main/kotlin/org/usvm/machine/types/VirtualTypes.kt
```diff
@@ -40,7 +41,12 @@ data class ConcreteTypeNegation(val concreteType: ConcretePythonType) : VirtualP
     }
 }
 
-sealed class TypeProtocol : VirtualPythonType() {
+/*
+ * Temporary slotId is nullable,
+ * since some slots, such as nb_int, nb_index and mp_length,
+ * are missing their implementation in virtual_objects.c
+ */
+sealed class TypeProtocol(val slotId: SlotId? = null) : VirtualPythonType() {
```
sergeypospelov, line 49: Why is `slotId` not used?

usvm-python/usvm-python-annotations/src/main/kotlin/org/usvm/annotations/ids/SlotId.kt
```diff
@@ -0,0 +1,70 @@
...
+    ;
+
+    companion object {
+        init {
+            values().filter { !it.mandatory }.forEachIndexed { index, entry ->
+                entry.maskBit = index
+            }
+        }
+    }
+    private var maskBit: Int? = null
```
sergeypospelov, line 46: IMO, it looks weird. Why can't we use ordinal and just ignore some bits?

usvm-python/usvm-python-annotations/src/main/kotlin/org/usvm/annotations/ids/SlotId.kt
```diff
@@ -0,0 +1,70 @@
...
+    fun getMaskBit(): Int = maskBit ?: error("No bits in the mask correspond to a mandatory slot.")
+}
+
+fun ByteArray.swapSlotBit(slot: SlotId): ByteArray {
```
sergeypospelov, line 50: Why is it not used?

tochilinak:
General comment for @sergeypospelov. This is some preliminary work for fixing a major issue: `MockType` now supports only all type protocols all together, so it cannot satisfy some constraint like `isSubtype(HasNbAdd).not()`. To be able to fix this, we need to be able to create virtual objects with different sets of slots. This PR provides an API to do that. Using it for fixing the initial problem is the next step.

tochilinak (approved):

sergeypospelov (approved):

PR #215, opened 2024-09-30
Basic USVM TS type system with type coercion

zishkaz:


usvm-core/src/main/kotlin/org/usvm/StateForker.kt
```diff
@@ -11,7 +12,39 @@ private typealias StateToCheck = Boolean
 private const val ForkedState = true
 private const val OriginalState = false
 
-sealed interface StateForker {
+interface StateForker {
```
CaelmBleidd, line 15: We don't need a separate TsForker. Seems like the only difference is that it works with UJoinedExpression, but it is a common API, and can be extracted into the original forker

usvm-core/src/main/kotlin/org/usvm/StateForker.kt
```diff
@@ -254,41 +287,12 @@ object NoSolverStateForker : StateForker {
...
-    return SplittedModels(trueModels, falseModels, unknownModels)
-}
-
 data class ForkResult<T>(
     val positiveState: T?,
     val negativeState: T?,
 )
 
-private data class SplittedModels<Type>(
+data class SplittedModels<Type>(
```
CaelmBleidd, line 295: Guess this will be private again

usvm-ts/src/main/kotlin/org/usvm/TSBinaryOperator.kt
```diff
@@ -6,32 +6,93 @@ sealed class TSBinaryOperator(
     val onBool: TSContext.(UExpr<UBoolSort>, UExpr<UBoolSort>) -> UExpr<out USort> = shouldNotBeCalled,
     val onBv: TSContext.(UExpr<UBvSort>, UExpr<UBvSort>) -> UExpr<out USort> = shouldNotBeCalled,
     val onFp: TSContext.(UExpr<UFpSort>, UExpr<UFpSort>) -> UExpr<out USort> = shouldNotBeCalled,
+    val onRef: TSContext.(UExpr<UAddressSort>, UExpr<UAddressSort>) -> UExpr<out USort> = shouldNotBeCalled,
```
CaelmBleidd, line 9: Please, add comments about desired sorts and ban sorts here, it is not obvious

usvm-ts/src/main/kotlin/org/usvm/TSBinaryOperator.kt
```diff
@@ -6,32 +6,93 @@ sealed class TSBinaryOperator(
...
+    val desiredSort: TSContext.(USort, USort) -> USort = { _, _ -> error("Should not be called") },
+    val banSorts: TSContext.(UExpr<out USort>, UExpr<out USort>) -> Set<USort> = {_, _ -> emptySet() },
 ) {
 
     object Eq : TSBinaryOperator(
         onBool = UContext<TSSizeSort>::mkEq,
         onBv = UContext<TSSizeSort>::mkEq,
         onFp = UContext<TSSizeSort>::mkFpEqualExpr,
+        onRef = UContext<TSSizeSort>::mkEq,
```
CaelmBleidd, line 18: Why not `mkHeapRefEq`?

usvm-ts/src/main/kotlin/org/usvm/TSBinaryOperator.kt
```diff
@@ -6,32 +6,93 @@ sealed class TSBinaryOperator(
...
+    // "a (untyped arg) != 1.0 (fp64 number)"
+    // can't yield
+    // "a (bool reg reading) != true", since "1.0.toBool() = true" is a new value for 1.0.
+    //
+    // So, that's the reason why banSorts in Neq throws out all primitive types except one of the expressions' one.
     object Neq : TSBinaryOperator(
         onBool = { lhs, rhs -> lhs.neq(rhs) },
         onBv = { lhs, rhs -> lhs.neq(rhs) },
         onFp = { lhs, rhs -> mkFpEqualExpr(lhs, rhs).not() },
+        onRef = { lhs, rhs -> lhs.neq(rhs) },
```
CaelmBleidd, line 33: Same

usvm-ts/src/main/kotlin/org/usvm/TSBinaryOperator.kt
```diff
@@ -6,32 +6,93 @@ sealed class TSBinaryOperator(
...
+        desiredSort = { lhs, _ -> lhs }
     )
 
+    // Neq must not be applied to a pair of expressions generated while initial Neq application.
+    // For example,
+    // "a (untyped arg) != 1.0 (fp64 number)"
+    // can't yield
+    // "a (bool reg reading) != true", since "1.0.toBool() = true" is a new value for 1.0.
+    //
+    // So, that's the reason why banSorts in Neq throws out all primitive types except one of the expressions' one.
```
CaelmBleidd, line 28: I didn't understand what it means

usvm-ts/src/main/kotlin/org/usvm/TSBinaryOperator.kt
```diff
@@ -6,32 +6,93 @@ sealed class TSBinaryOperator(
...
         onBv = { lhs, rhs -> lhs.neq(rhs) },
         onFp = { lhs, rhs -> mkFpEqualExpr(lhs, rhs).not() },
+        onRef = { lhs, rhs -> lhs.neq(rhs) },
+        desiredSort = { lhs, _ -> lhs },
+        banSorts = { lhs, rhs ->
+            when {
+                lhs is TSWrappedValue ->
+                    // rhs.sort == addressSort is a mock not to cause undefined
+                    // behaviour with support of new language features.
+                    if (rhs is TSWrappedValue || rhs.sort == addressSort) emptySet() else
```
CaelmBleidd, line 40: Please, use `{}` and another formatting. 

`if () { 
} else {
}
`
The only exception are one-liners

CaelmBleidd, line 40: Why addresses are not supported? I didn't get it. If some feature is not implemented it should be marked as "TODO", not to work in incorrect way

usvm-ts/src/main/kotlin/org/usvm/TSBinaryOperator.kt
```diff
@@ -6,32 +6,93 @@ sealed class TSBinaryOperator(
...
+                        .minus(lhs.sort)
+                else -> emptySet()
+            }
+        }
+    )
+
+    object Add : TSBinaryOperator(
+        onFp = { lhs, rhs -> mkFpAddExpr(fpRoundingModeSortDefaultValue(), lhs, rhs) },
+        onBv = UContext<TSSizeSort>::mkBvAddExpr,
+        desiredSort = { _, _ -> fp64Sort },
```
CaelmBleidd, line 58: Same here, it will be very difficult to find such places without todo or errors

usvm-ts/src/main/kotlin/org/usvm/TSBinaryOperator.kt
```diff
@@ -6,32 +6,93 @@ sealed class TSBinaryOperator(
...
+    )
+
+    internal operator fun invoke(lhs: UExpr<out USort>, rhs: UExpr<out USort>, scope: TSStepScope): UExpr<out USort> {
+        val bannedSorts = lhs.tctx.banSorts(lhs, rhs)
+
+        fun apply(lhs: UExpr<out USort>, rhs: UExpr<out USort>): UExpr<out USort>? {
+            val ctx = lhs.tctx
+            val lhsSort = lhs.sort
+            val rhsSort = rhs.sort
+            assert(lhsSort == rhsSort)
```
CaelmBleidd, line 73: Лучше не использовать джавовские асёрты. Во-первых, они могут быть отключены. Во-вторых, они дорогие

usvm-ts/src/main/kotlin/org/usvm/TSBinaryOperator.kt
```diff
@@ -6,32 +6,93 @@ sealed class TSBinaryOperator(
...
+
+            if (lhsSort in bannedSorts) return null
+            if (ctx.desiredSort(lhsSort, rhsSort) != lhsSort) return null
```
CaelmBleidd, line 76: These places require clarifying comments as well

usvm-ts/src/main/kotlin/org/usvm/TSBinaryOperator.kt
```diff
@@ -6,32 +6,93 @@ sealed class TSBinaryOperator(
...
+        }
+
         val lhsSort = lhs.sort
         val rhsSort = rhs.sort
 
-        if (lhsSort != rhsSort) TODO("Implement type coercion")
+        val ctx = lhs.tctx
+        val sort = ctx.desiredSort(lhsSort, rhsSort)
 
         return when {
```
CaelmBleidd, line 93: There are no other cases?

usvm-ts/src/main/kotlin/org/usvm/TSExprTransformer.kt
```diff
@@ -0,0 +1,149 @@
...
+import org.jacodb.ets.base.EtsType
+
+typealias CoerceAction = (UExpr<out USort>, UExpr<out USort>) -> UExpr<out USort>?
+
+class TSExprTransformer(
+    private val baseExpr: UExpr<out USort>,
+    private val scope: TSStepScope,
+) {
+
+    private val exprCache: MutableMap<USort, UExpr<out USort>?> = mutableMapOf(baseExpr.sort to baseExpr)
```
CaelmBleidd, line 17: If you don't need order, use `hashMapOf` instead

usvm-ts/src/main/kotlin/org/usvm/TSExprTransformer.kt
```diff
@@ -0,0 +1,149 @@
...
+            TSTypeSystem.primitiveTypes.forEach { transform(ctx.typeToSort(it), modifyConstraints = false) }
+        }
+    }
+
+    fun transform(sort: USort, modifyConstraints: Boolean = true): UExpr<out USort>? = with(ctx) {
+        val (result, type) = when (sort) {
+            fp64Sort -> asFp64() to EtsNumberType
+            boolSort -> asBool() to EtsBooleanType
+            addressSort -> asRef() to null
+            else -> error("")
```
CaelmBleidd, line 31: Empty error message

usvm-ts/src/main/kotlin/org/usvm/TSExprTransformer.kt
```diff
@@ -0,0 +1,149 @@
...
+        if (baseExpr.sort == ctx.addressSort) {
+            TSTypeSystem.primitiveTypes.forEach { transform(ctx.typeToSort(it), modifyConstraints = false) }
+        }
+    }
+
+    fun transform(sort: USort, modifyConstraints: Boolean = true): UExpr<out USort>? = with(ctx) {
+        val (result, type) = when (sort) {
+            fp64Sort -> asFp64() to EtsNumberType
+            boolSort -> asBool() to EtsBooleanType
+            addressSort -> asRef() to null
```
CaelmBleidd, line 30: Please add a comment about this null

usvm-ts/src/main/kotlin/org/usvm/TSExprTransformer.kt
```diff
@@ -0,0 +1,149 @@
...
+        action: CoerceAction
+    ): UExpr<out USort> {
+        intersect(other)
+
+        val rawExprs = exprCache.keys.mapNotNull { sort ->
+            val lhv = transform(sort)
+            val rhv = other.transform(sort)
+            if (lhv != null && rhv != null) {
+                action(lhv, rhv)
+            } else null
```
CaelmBleidd, line 50: {}

usvm-ts/src/main/kotlin/org/usvm/TSExprTransformer.kt
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        val innerCoercionExprs = this.generateAdditionalExprs(rawExprs) + other.generateAdditionalExprs(rawExprs)
+
+        val exprs = rawExprs + innerCoercionExprs
+
+        return if (exprs.size > 1) {
+            assert(exprs.all { it.sort == ctx.boolSort })
+            UJoinedBoolExpr(ctx, exprs.cast())
+        } else exprs.single()
```
CaelmBleidd, line 60: {}

CaelmBleidd, line 60: А если там ничего не будет? Если так не бывает, то лучше использовать singleOrNull и повернуть условия в другую сторону

usvm-ts/src/main/kotlin/org/usvm/TSExprTransformer.kt
```diff
@@ -0,0 +1,149 @@
...
+     * Generates and caches additional constraints for coercion expression list.
+     *
+     * For now used to save link between fp and bool sorts of [baseExpr].
+     *
+     * @return List of additional [UExpr].
+     */
+    private fun generateAdditionalExprs(rawExprs: List<UExpr<out USort>>): List<UExpr<out USort>> = with(ctx) {
+        if (!rawExprs.all { it.sort == boolSort }) return emptyList()
+        val newExpr = when (baseExpr.sort) {
+            addressSort -> addedExprCache.putOrNull(mkEq(fpToBoolSort(asFp64()), asBool()))
```
CaelmBleidd, line 84: This is unclear, why we add info about bv bool transformation if we have an address

usvm-ts/src/main/kotlin/org/usvm/TSExprTransformer.kt
```diff
@@ -0,0 +1,149 @@
...
+
+    fun asFp64(): UExpr<KFp64Sort> = exprCache.getOrPut(ctx.fp64Sort) {
+        when (baseExpr.sort) {
+            ctx.fp64Sort -> baseExpr
+            ctx.boolSort -> ctx.boolToFpSort(baseExpr.cast())
+            ctx.addressSort -> with(ctx) {
+                TSRefTransformer(ctx, fp64Sort).apply(baseExpr.cast()).cast()
+            }
+
+            else -> ctx.mkFp64(0.0)
```
CaelmBleidd, line 99: Why?

usvm-ts/src/main/kotlin/org/usvm/TSExprTransformer.kt
```diff
@@ -0,0 +1,149 @@
...
+
+    fun asBool(): UExpr<KBoolSort> = exprCache.getOrPut(ctx.boolSort) {
+        when (baseExpr.sort) {
+            ctx.boolSort -> baseExpr
+            ctx.fp64Sort -> ctx.fpToBoolSort(baseExpr.cast())
+            ctx.addressSort -> with(ctx) {
+                TSRefTransformer(ctx, boolSort).apply(baseExpr.cast()).cast()
+            }
+
+            else -> ctx.mkFalse()
```
CaelmBleidd, line 112: Why? If something is not implemented or unexpected it should throw an error

usvm-ts/src/main/kotlin/org/usvm/TSExpressions.kt
```diff
@@ -34,6 +36,95 @@ class TSUndefinedValue(ctx: TSContext) : UExpr<TSUndefinedSort>(ctx) {
...
+class UJoinedBoolExpr(
+    ctx: TSContext,
+    val exprs: List<UBoolExpr>
+) : UBoolExpr(ctx) {
+    override val sort: UBoolSort
+        get() = ctx.boolSort
+
+    private val joinedExprs = ctx.mkAnd(exprs)
+
+    fun not(): UBoolExpr = ctx.mkAnd(exprs.map(ctx::mkNot))
```
CaelmBleidd, line 53: Are these conditions big? Should we consider storing these negations inside it?

usvm-ts/src/main/kotlin/org/usvm/TSExpressions.kt
```diff
@@ -34,6 +36,95 @@ class TSUndefinedValue(ctx: TSContext) : UExpr<TSUndefinedSort>(ctx) {
...
+        return coerce(other, action)
+    }
+
+    override fun accept(transformer: KTransformerBase): KExpr<USort> {
+        return value.cast()
+    }
+
+    override fun internEquals(other: Any): Boolean = structurallyEqual(other)
+
+    override fun internHashCode(): Int = hash()
```
CaelmBleidd, line 118: Why these methods are not linked with a value?

usvm-ts/src/main/kotlin/org/usvm/TSInterpreter.kt
```diff
@@ -77,7 +77,7 @@ class TSInterpreter(
         val exprResolver = exprResolverWithScope(scope)
 
         val boolExpr = exprResolver
-            .resolveTSExpr(stmt.condition)
+            .resolveTSExprNoUnwrap(stmt.condition)
```
CaelmBleidd, line 80: Why we don't need to unwrap it?

usvm-ts/src/main/kotlin/org/usvm/TSTypeStorage.kt
```diff
@@ -0,0 +1,28 @@
...
+    TODO: support other possibly untyped refs.
+ */
+class TSTypeStorage(
+    private val ctx: TSContext,
+    private val keyToTypes: MutableMap<Any, MutableSet<EtsType>> = mutableMapOf(),
+) {
+
+    fun storeSuggestedType(ref: UExpr<UAddressSort>, type: EtsType) {
+        // TODO: finalize implementation and remove this assert
+        assert(ref is URegisterReading)
```
CaelmBleidd, line 18: assert

usvm-ts/src/main/kotlin/org/usvm/TSTypeStorage.kt
```diff
@@ -0,0 +1,28 @@
+package org.usvm
+
+import org.jacodb.ets.base.EtsType
+
+/*
+    This is a very basic implementation of type storage with memory and model objects interoperability.
```
CaelmBleidd, line 6: How it works? For the same state there are multiple possible models available. Why is it usefull?

usvm-ts/src/main/kotlin/org/usvm/TSTypeSystem.kt
```diff
@@ -1,37 +1,136 @@
...
+                allTypes.toTypesResult(wasTimeoutExpired = true)
+            }
+        }
+    }
+
+    override val isEmpty: Boolean?
+        get() = anyTypeStream.isEmpty?.let { primitiveTypes.isEmpty() }
+
+    override val commonSuperType: EtsType?
+        get() = EtsUnknownType.takeIf { !(isEmpty ?: true) }
```
CaelmBleidd, line 133: Please add todo to such places

usvm-ts/src/main/kotlin/org/usvm/TSUnaryOperator.kt
```diff
@@ -0,0 +1,36 @@
...
+) {
+
+    object Not : TSUnaryOperator(
+        onBool = TSContext::mkNot,
+        desiredSort = { boolSort },
+    )
+
+    internal operator fun invoke(operand: UExpr<out USort>, scope: TSStepScope): UExpr<out USort> = with(operand.tctx) {
+        val sort = this.desiredSort(operand.sort)
+        val expr = if (operand is TSWrappedValue) operand.asSort(sort) else
```
CaelmBleidd, line 20: {}

usvm-ts/src/main/kotlin/org/usvm/TSUnaryOperator.kt
```diff
@@ -0,0 +1,36 @@
...
+            TSExprTransformer(operand, scope).transform(sort)
+
+        when (expr?.sort) {
+            is UBoolSort -> onBool(expr.cast())
+            is UBvSort -> onBv(expr.cast())
+            is UFpSort -> onFp(expr.cast())
+            null -> mkNullRef()
```
CaelmBleidd, line 27: Why?

usvm-ts/src/test/kotlin/org/usvm/util/TSMethodTestRunner.kt
```diff
@@ -172,6 +173,8 @@ open class TSMethodTestRunner : TestRunner<TSTest, MethodDescriptor, EtsType?, T
                 TSObject.TSNumber.Double::class -> EtsNumberType
                 TSObject.TSNumber.Integer::class -> EtsNumberType
                 TSObject.UndefinedObject::class -> EtsUndefinedType
+                TSObject.Object::class -> EtsUnknownType
+                TSObject::class -> EtsUnknownType
```
CaelmBleidd, line 177: Look suspicious

usvm-ts/src/test/kotlin/org/usvm/util/TSMethodTestRunner.kt
```diff
@@ -204,15 +207,15 @@ open class TSMethodTestRunner : TestRunner<TSTest, MethodDescriptor, EtsType?, T
             val fileURL = javaClass.getResource("/samples/${packagePath}/${id.fileName}")
                 ?: error("No such file found")
             val filePath = Paths.get(fileURL.toURI())
-            val file = loadEtsFileAutoConvert(filePath)
+            val scene = EtsScene(listOf(loadEtsFileAutoConvert(filePath)))
 
-            val method = file.getMethodByDescriptor(id)
+            val method = scene.files.single().getMethodByDescriptor(id)
```
CaelmBleidd, line 212: I'm not sure that it always contains only a single file

usvm-ts/src/test/kotlin/org/usvm/util/TSTestResolver.kt
```diff
@@ -10,32 +11,38 @@ import org.jacodb.ets.base.EtsRefType
...
+import org.usvm.extractOrThis
 import org.usvm.memory.URegisterStackLValue
+import org.usvm.model.UModelBase
 import org.usvm.state.TSMethodResult
 import org.usvm.state.TSState
+import org.usvm.types.first
 
-class TSTestResolver {
+class TSTestResolver(
+    private val state: TSState
```
CaelmBleidd, line 36: Why? Resolver should be able to resolve many states independently

usvm-ts/src/test/kotlin/org/usvm/util/TSTestResolver.kt
```diff
@@ -52,10 +59,48 @@ class TSTestResolver {
...
+        ctx: TSContext,
+        model: UModelBase<EtsType>,
+    ): List<TSObject> = with(ctx) {
+        params.mapIndexed { idx, param ->
+            val type = param.type
+            val lValue = URegisterStackLValue(typeToSort(type), idx)
+            val expr = model.read(lValue).extractOrThis()
+            if (type is EtsUnknownType) {
+                approximateParam(expr.cast(), idx, model)
+            } else resolveExpr(expr, type, model)
```
CaelmBleidd, line 73: {}

usvm-ts/src/test/kotlin/org/usvm/util/TSTestResolver.kt
```diff
@@ -52,10 +59,48 @@ class TSTestResolver {
...
+        }
+    }
+
+    private fun approximateParam(expr: UConcreteHeapRef, idx: Int, model: UModelBase<EtsType>): TSObject =
+        with(expr.ctx as TSContext) {
+            val suggestedType = state.getSuggestedType(idx)
+            return suggestedType?.let { newType ->
```
CaelmBleidd, line 80: Something odd is happening here. You can always access all required information using symbolic values and symbolic values only, otherwise you will encounter path divergence. This map **cannot** be built from concrete values to type info, only from symbolic ones

CaelmBleidd (changes requested):

.github/workflows/build-and-run-tests.yml
```diff
@@ -40,24 +40,24 @@ jobs:
           DEST_DIR="arkanalyzer"
           MAX_RETRIES=10
           RETRY_DELAY=3  # Delay between retries in seconds  
-          BRANCH="neo/2024-08-07"
```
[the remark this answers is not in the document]
zishkaz: Set as version in jacodb/neo branch, from which we currently get jacodb artifact for usvm

usvm-ts/src/main/kotlin/org/usvm/TSBinaryOperator.kt
zishkaz: No, since chosen sort is only used to intersect both exprCaches and have at least one sort to apply binary operation to

usvm-ts/src/main/kotlin/org/usvm/TSExprTransformer.kt
zishkaz: По построению не может быть меньше 1 элемента, чтобы не писать большой код с кучей проверок пускай .single() бросает эксепшн -- это значит что неверно построился список ```exprs```

usvm-ts/src/main/kotlin/org/usvm/TSExpressions.kt
zishkaz: No, they are even theoretically small and this function is rarely used. A broader explanation will be inlined in the code above.

usvm-ts/src/main/kotlin/org/usvm/TSInterpreter.kt
zishkaz: We don't want to lose possible UJoinedBoolExpr for futher fork here

usvm-ts/src/main/kotlin/org/usvm/TSExpressions.kt
zishkaz: I just ripped off these implementations from some core usvm symbols, will mark it as draft with a comment

usvm-ts/src/main/kotlin/org/usvm/TSTypeSystem.kt
zishkaz: These are finished implementations

usvm-ts/src/main/kotlin/org/usvm/TSBinaryOperator.kt
zishkaz: Done

usvm-ts/src/main/kotlin/org/usvm/TSBinaryOperator.kt
zishkaz: Fixed

usvm-ts/src/main/kotlin/org/usvm/TSBinaryOperator.kt
zishkaz: Fixed

usvm-ts/src/main/kotlin/org/usvm/TSBinaryOperator.kt
zishkaz: Enriched explanation

usvm-ts/src/main/kotlin/org/usvm/TSBinaryOperator.kt
zishkaz: Enriched explanation

usvm-ts/src/main/kotlin/org/usvm/TSExprTransformer.kt
zishkaz: Fixed

usvm-ts/src/main/kotlin/org/usvm/TSExprTransformer.kt
zishkaz: Added

usvm-ts/src/main/kotlin/org/usvm/TSExprTransformer.kt
zishkaz: Explanation added

usvm-ts/src/main/kotlin/org/usvm/TSExprTransformer.kt
zishkaz: Fixed

usvm-ts/src/main/kotlin/org/usvm/TSExprTransformer.kt
zishkaz: Fixed

usvm-ts/src/main/kotlin/org/usvm/TSTypeStorage.kt
zishkaz: removed

usvm-ts/src/main/kotlin/org/usvm/TSTypeStorage.kt
zishkaz: For different entities (register readings, lists, etc.) with ```addressSort``` it stores type suggestions from type coercion. Different models will have the same access path to their types as each type of ref entity has its own way of defining a key.

zishkaz: For example, registet readings use its' ```idx``` as a key in TypeStorage

zishkaz: Explanation enriched in the code, please check

usvm-ts/src/test/kotlin/org/usvm/util/TSMethodTestRunner.kt
zishkaz: It's alright, ```TSObject::Object``` is typescript obj representation. ```TSObject``` is for untyped tests, since we don't want to limit which objects we expect to serialize from models because of type coercion.

usvm-ts/src/test/kotlin/org/usvm/util/TSMethodTestRunner.kt
zishkaz: Yeah, but this is just a jacodb API update fixup. Currently I don't want to write a correct search for a method among all scene files.

usvm-ts/src/main/kotlin/org/usvm/TSUnaryOperator.kt
zishkaz: Fixed

usvm-ts/src/test/kotlin/org/usvm/util/TSTestResolver.kt
zishkaz: ```TSTestResolver``` was instantiated for each returned state, and I needed access to ```state.typeStorage```.
So I decided to add state as private constructor param.

usvm-ts/src/test/kotlin/org/usvm/util/TSTestResolver.kt
zishkaz: I believe different types of ref entities (reg readings, etc.) have unique ways to generate a key. For example, reg readings can use its' ```idx``` field as a key, and this key leads to correct list of types from any model.

usvm-ts/src/main/kotlin/org/usvm/TSExprTransformer.kt
CaelmBleidd: Why don't we need to add a connection between baseSort expr and them? If it already exists for baseSort <-> fp and bool, it exists for bool and fp by transitivity as well

usvm-ts/src/test/kotlin/org/usvm/util/TSMethodTestRunner.kt
CaelmBleidd: It is a sealed interface, how we have a situation when we have something different from already been analyzed? For me it looks like an error

usvm-ts/src/test/kotlin/org/usvm/util/TSTestResolver.kt
CaelmBleidd: But you might store it with another key -- if I resolve something from registerReading(1), it doesn't mean that I always accessed this value using this key. Probably, I could do it with registerReading(2).getByIndex(3), i could have accessed it with RR(3) (because of aliasing) and many many other ways

CaelmBleidd (dismissed):

Lipen (approved):
Very good!

usvm-ts/src/main/kotlin/org/usvm/machine/interpreter/TSInterpreter.kt
```diff
@@ -89,11 +89,16 @@ class TSInterpreter(
...
-            }
+        val conditionExpr = exprResolver.resolve(stmt.condition) ?: run {
+            logger.warn { "Failed to resolve condition: $stmt" }
+            return
+        }
+
+        val boolExpr = if (conditionExpr.sort == ctx.boolSort) {
+            conditionExpr.asExpr(ctx.boolSort)
+        } else {
+            ctx.mkTruthyExpr(conditionExpr, scope)
```
CaelmBleidd, line 100: Is it truly `truthy` or it will be reversed as the condition? For example, not `if (x)`, but `if (!x)`. If we sure that `!` forces value to be boolean, is it a reachable branch?

Lipen: `!` indeed seems to be forcing value to become `boolean`, however not all if-stmts have their conditions reversed. It might be a general observed behavior of the current state of ArkIR and ABC (so that this `mkTruthy` branch is indeed unreachable for "real" code), yet nobody can guarantee it. Moreover, non-negated (plain `if (x)`) conditions arise from the manually constructed IR, like in the current tests (and I do believe it is the correct way here, since it allows for more predictable testing), that's why I even implemented this branch to avoid mismatched sort here.

Overall, it does not hurt anybody to have this "correct" truthy-based impl here, even if it will be "unreachable" on the "real" code.

CaelmBleidd: My main concern is about consistency of conditions and branches, so we need to be sure that with `if (x)` condition we'll have the correct `positive` and `negative` branches

Lipen: Both `conditionExpr.asExpr(ctx.boolSort)` and `ctx.mkTruthyExpr(conditionExpr, scope)` represent "positive" branch. And both of them have `boolSort`, so the "negative" branch is going to be a simple negation of it. Isn't it?

usvm-ts/src/main/kotlin/org/usvm/machine/TSContext.kt
```diff
@@ -0,0 +1,193 @@
...
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.contract
+
+typealias TSSizeSort = UBv32Sort
+
+class TSContext(
+    val scene: EtsScene,
+    components: TSComponents,
+) : UContext<TSSizeSort>(components) {
+    val undefinedSort: TSUndefinedSort by lazy { TSUndefinedSort(this) }
```
Saloed, line 39: Do we really need undefined sort?

usvm-ts/src/main/kotlin/org/usvm/machine/TSContext.kt
```diff
@@ -0,0 +1,193 @@
...
+
+    val unresolvedSort: TSUnresolvedSort = TSUnresolvedSort(this)
+
+    private val undefinedValue by lazy { TSUndefinedValue(this) }
```
Saloed, line 43: Mybe we can use null?

usvm-ts/src/main/kotlin/org/usvm/machine/TSContext.kt
```diff
@@ -0,0 +1,193 @@
...
+    fun typeToSort(type: EtsType): USort = when (type) {
+        is EtsBooleanType -> boolSort
+        is EtsNumberType -> fp64Sort
+        is EtsRefType -> addressSort
+        is EtsUnknownType -> unresolvedSort
+        else -> TODO("Support all JacoDB types, encountered $type")
+    }
+
+    // TODO fix conjuncts
+    fun mkTruthyExpr(expr: UExpr<out USort>, scope: TSStepScope): UBoolExpr = scope.calcOnState {
```
Saloed, line 54: I don't like logic with `TSStepScope` inside the Context class. Maybe it is better to move it somewhere else, e.g. as extension function

usvm-ts/src/main/kotlin/org/usvm/machine/expr/TSExprResolver.kt
```diff
@@ -0,0 +1,555 @@
...
+        return resolveBinaryOperator(TSBinaryOperator.And, expr)
+    }
+
+    override fun visit(expr: EtsNotEqExpr): UExpr<out USort>? {
+        return resolveBinaryOperator(TSBinaryOperator.Neq, expr)
+    }
+
+    override fun visit(expr: EtsNotExpr): UExpr<out USort>? = with(ctx) {
+        resolveAfterResolved(expr.arg) { arg ->
+            // TSUnaryOperator.Not(arg, scope)
```
Saloed, line 176: Why commented?

usvm-ts/src/main/kotlin/org/usvm/machine/expr/TSExprResolver.kt
```diff
@@ -0,0 +1,555 @@
...
+        error("Not supported $expr")
+    }
+
+    override fun visit(expr: EtsMulExpr): UExpr<out USort>? {
+        logger.warn { "visit(${expr::class.simpleName}) is not implemented yet" }
+        error("Not supported $expr")
+    }
+
+    override fun visit(expr: EtsNegExpr): UExpr<out USort>? {
+        if (expr.type is EtsNumberType) {
```
Saloed, line 291: Maybe move to unary operator?

usvm-ts/src/main/kotlin/org/usvm/machine/expr/TSExprResolver.kt
```diff
@@ -0,0 +1,555 @@
...
+
+    override fun visit(value: EtsArrayAccess): UExpr<out USort>? {
+        logger.warn { "visit(${value::class.simpleName}) is not implemented yet" }
+        error("Not supported $value")
+    }
+
+    override fun visit(value: EtsInstanceFieldRef): UExpr<out USort>? = with(ctx) {
+        val instanceRef = resolve(value.instance)?.asExpr(addressSort) ?: return null
+        // TODO: checkNullPointer(instanceRef)
+        val fieldType = scene.projectAndSdkClasses
```
Saloed, line 416: Let's move this field type lookup somewhere else. This thing will be a reason of a huge performance issues. So, it would be nice if we can easily detect it.

usvm-ts/src/main/kotlin/org/usvm/machine/interpreter/TSInterpreter.kt
```diff
@@ -111,18 +126,26 @@ class TSInterpreter(
...
+            memory.write(lValue, expr.asExpr(lValue.sort), guard = ctx.trueExpr)
+
             val nextStmt = stmt.nextStmt ?: return@doWithState
             newStmt(nextStmt)
         }
     }
 
     private fun visitCallStmt(scope: TSStepScope, stmt: EtsCallStmt) {
-        TODO()
+        TODO() // IMPORTANT do not forget to fill sorts of arguments map
```
Saloed, line 148: This is also relevant for the call `expr`.

usvm-ts/src/main/kotlin/org/usvm/machine/operator/TSBinaryOperator.kt
```diff
@@ -0,0 +1,511 @@
...
+                        expr = mkFpEqualExpr(
+                            memory.read(getIntermediateFpLValue(lhs.address)),
+                            boolToFp(memory.read(getIntermediateBoolLValue(rhs.address)))
+                        )
+                    )
+
+                    // TODO: support objects
+                }
+
+                if (lhs.isFakeObject()) {
```
Saloed, line 167: This condition always holds when the `lhs.isFakeObject() && rhs.isFakeObject()` holds. So, we need return or smth

usvm-ts/src/main/kotlin/org/usvm/machine/state/TSState.kt
```diff
@@ -26,6 +27,7 @@ class TSState(
     forkPoints: PathNode<PathNode<EtsStmt>> = PathNode.root(),
     var methodResult: TSMethodResult = TSMethodResult.NoCall,
     targets: UTargetsSet<TSTarget, EtsStmt> = UTargetsSet.empty(),
+    private val localToSort: MutableMap<EtsMethod, MutableMap<Int, USort>> = hashMapOf(),
```
Saloed, line 30: Maybe it is better to use persistent map?

Saloed (changes requested):

usvm-ts/src/main/kotlin/org/usvm/machine/expr/TSExprResolver.kt
CaelmBleidd: Added todo, will move when we support all unary operators

usvm-ts/src/main/kotlin/org/usvm/machine/state/TSState.kt
CaelmBleidd: I think it's ok for now due to multiple potential put actions during getOrPutSoertForLocal methods

usvm-ts/src/main/kotlin/org/usvm/machine/TSContext.kt
CaelmBleidd: Yes, I think we need to rewrite it using Undefined as null in other languages, but for now it is too complicated to think about. I suggest to do it later, when support of undefined will be added

usvm-ts/src/main/kotlin/org/usvm/machine/state/TSState.kt
Saloed: You can use `UPersistentHashMap` which is designed to perform write operations. You can find usage examples in `UMemory` (`regions` property)

usvm-ts/src/main/kotlin/org/usvm/machine/TSContext.kt
```diff
@@ -0,0 +1,126 @@
...
+        scope: TSStepScope,
+        boolValue: UBoolExpr? = null,
+        fpValue: UExpr<KFp64Sort>? = null,
+        refValue: UHeapRef? = null,
+    ): UConcreteHeapRef {
+        require(boolValue != null || fpValue != null || refValue != null) {
+            "Fake object should contain at least one value"
+        }
+
+        return scope.calcOnState {
```
Saloed, line 72: Another one scope operation in context

PR #217, opened 2024-10-18
Fixes for bugs found during approximations testing

IgorFilimonov:
Fixes bugs such as:

1. Incorrect size of a list after addAll()
2. Inability of maps to hold nulls as keys (for example, appropriate map.put(null, null) used to throw "nullRef on top")
3. Object cloning
4. Now test for ArrayList decoder should be passed


usvm-core/src/main/kotlin/org/usvm/api/collection/ListCollectionApi.kt
```diff
@@ -1,22 +1,12 @@
 package org.usvm.api.collection
 
-import org.usvm.StepScope
-import org.usvm.UContext
-import org.usvm.UExpr
-import org.usvm.UHeapRef
-import org.usvm.USort
-import org.usvm.UState
+import org.usvm.*
```
Saloed, line 3: Please, avoid such changes

usvm-jvm/src/main/kotlin/org/usvm/machine/JcApproximations.kt
```diff
@@ -89,6 +77,12 @@ class JcMethodApproximationResolver(
     private val usvmApiSymbolicMap by lazy { ctx.cp.findClassOrNull<SymbolicMap<*, *>>() }
     private val usvmApiSymbolicIdentityMap by lazy { ctx.cp.findClassOrNull<SymbolicIdentityMap<*, *>>() }
 
+    @Suppress("RecursivePropertyAccessor")
```
Saloed, line 80: This appears to be an Intellij warning, not a Kotlin compiler warning. Please, don't suppress such warnings.

Saloed:
```
ApproximationsDecoderTest > testArrayListDecoder() FAILED
    java.lang.NullPointerException at ApproximationsDecoderTest.kt:760
```

Saloed:
Will be reopened later

PR #223, opened 2024-12-09
Generalize `JcDebugProfileObserver` for `usvm-core`

tochilinak:


usvm-core/src/main/kotlin/org/usvm/statistics/UDebugProfileObserver.kt
```diff
@@ -0,0 +1,289 @@
...
+        /**
+         * Return null if statement is not a call statement
+         * */
+        fun getMethodToCallIfCallStatement(statement: Statement): Method?
+
+        // operations with default implementations
+
+        fun printStatement(statement: Statement): String = statement.toString()
+        fun printMethodName(method: Method): String = method.toString()
+        fun getOriginalInst(statement: Statement): Statement = statement
```
Saloed, line 39: This is a jvm specific hack. So, let's add a comment about this to avoid confusion about what the original inst is.

usvm-core/src/main/kotlin/org/usvm/statistics/UDebugProfileObserver.kt
```diff
@@ -0,0 +1,289 @@
...
+                    val allStatements = statementOperations.getAllMethodStatements(method)
+                    allStatements.map { childStmt ->
+                        inst.find {
+                            it.key == childStmt
+                        }?.let {
+                            it.key to it.value
+                        } ?: let {
+                            val emptyFrame = ProfileFrame(
+                                childStmt,
+                                0L,
```
Saloed, line 213: It is better to use named args here

Saloed (changes requested):
Minor issues

Saloed (approved):

PR #244, opened 2025-02-05
Add ensureSat

Lipen:
This PR adds `USolverResult<T>.ensureSat(): USatResult<T>` extension function to replace "unsafe" casts `as USatResult<*>` with "safe" `.ensureSat()` calls.

### DIFF
--- a/usvm-core/src/main/kotlin/org/usvm/utils/ResultUtils.kt
+++ b/usvm-core/src/main/kotlin/org/usvm/utils/ResultUtils.kt
@@ -1,0 +1,9 @@
+package org.usvm.utils
+
+import org.usvm.solver.USatResult
+import org.usvm.solver.USolverResult
+
+fun <T> USolverResult<T>.assertSat(): USatResult<T> {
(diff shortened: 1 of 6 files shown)


usvm-core/src/main/kotlin/org/usvm/utils/ResultUtils.kt
```diff
@@ -0,0 +1,9 @@
+package org.usvm.utils
+
+import org.usvm.solver.USatResult
+import org.usvm.solver.USolverResult
+
+fun <T> USolverResult<T>.assertSat(): USatResult<T> {
```
Saloed, line 6: As for me, it is better to rename it as `ensureSat`. The reason, is that `assert` is closely related to SMT solver assert operation.

Saloed (changes requested):

usvm-core/src/main/kotlin/org/usvm/utils/ResultUtils.kt
Lipen: Fair enough

Saloed (approved):

PR #256, opened 2025-02-17
Add support for arrays

CaelmBleidd:
Required TODO:  
* Reading by an index not from the indices range (e.g., -1, 123132, 1.4, etc)
* Change the length of the array (both writing in non-existing index and methods with shift/unshift)
* Make it possible to write values of different types into a created homogeneous array, e.g. boolean in the array of numbers
* Errors for incorrect array creation, e.g, negative length, object as a length, etc.


Lipen (approved):
Good!

PR #258, opened 2025-02-18
Support arrays containing values of different types

CaelmBleidd:
Also, this request adds support for arrays expansion mechanism


Lipen (approved):
Very good! :rock:

PR #267, opened 2025-03-03
Add live variables analysis before type inference

MForest7:
- Add liveness analysis before type inference
  - `ForwardAnalyzer` stores facts about the variable immediately after the last instruction where the variable was used and `ForwardFlowFunction` drops facts about dead variables. This pre-analysis slightly increases the total performance without loss of precision.
- Fixed tests: now we doesn't take into account time spent to compute statistics
- Add flags for `ForwardAnalyzer` and `ForwardFlowFunctions` to enable/disable pre-analysis (alias and/or liveness)
- Add index to speed up callee resolver

### DIFF
--- a/usvm-ts-dataflow/src/main/kotlin/org/usvm/dataflow/ts/infer/Alias.kt
+++ b/usvm-ts-dataflow/src/main/kotlin/org/usvm/dataflow/ts/infer/Alias.kt
@@ -179,17 +179,17 @@ class StmtAliasInfoImpl(
         if (stmt !is EtsAssignStmt) {
             return this
         }
-        when (val rhv = stmt.rhv) {
+        return when (val rhv = stmt.rhv) {
             is EtsParameterRef -> {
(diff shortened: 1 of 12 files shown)


usvm-ts-dataflow/src/main/kotlin/org/usvm/dataflow/ts/infer/Alias.kt
```diff
@@ -494,14 +475,14 @@ class MethodAliasInfo(
         postOrderDfs(root)
         order.reverse()
 
-        fun computePreAliases(stmt: EtsStmt): StmtAliasInfo {
+        fun computePreAliases(stmt: EtsStmt): StmtAliasInfoImpl {
```
Lipen, line 478: You can keep less specific `StmtAliasInfo` return type.

Lipen:
@MForest7 thanks for your great work! :cn: :ramen: :panda_face:

PR #277, opened 2025-05-16
Add operator

CaelmBleidd:
This PR adds support for `Add` (`+`) operator for TS interpreter.

### DIFF
--- a/usvm-ts/src/main/kotlin/org/usvm/machine/operator/TsBinaryOperator.kt
+++ b/usvm-ts/src/main/kotlin/org/usvm/machine/operator/TsBinaryOperator.kt
@@ -569,7 +569,7 @@ sealed interface TsBinaryOperator {
                 }
             }
 
-            val loosyEqualityConstraint with(Eq) {
+            val loosyEqualityConstraint = with(Eq) {
                 resolveFakeObject(lhs, rhs, scope)
             }
 


usvm-ts/src/main/kotlin/org/usvm/machine/expr/ExprUtil.kt
```diff
@@ -125,5 +125,15 @@ fun TsContext.mkNumericExpr(
     // TODO: ToPrimitive, then ToNumber again
     // TODO: probably we need to implement Object (Ref/Fake) -> Number conversion here directly, without ToPrimitive
 
-    error("Unsupported sort: ${expr.sort}")
+    // TODO incorrect implementation, returns some number that is not equal to 0 and NaN
```
Lipen, line 128: Fix or create an issue

usvm-ts/src/main/kotlin/org/usvm/machine/expr/ExprUtil.kt
```diff
@@ -125,5 +125,15 @@ fun TsContext.mkNumericExpr(
...
+
+    return@calcOnState mkIte(
+        condition = mkEq(expr.asExpr(addressSort), mkTsNullValue()),
+        trueBranch = mkFp(0.0, fp64Sort),
+        falseBranch = mkIte(
+            mkEq(expr.asExpr(addressSort), mkUndefinedValue()),
+            mkFp64NaN(),
+            mkFp64NaN() // TODO we should not generalize it this way
```
Lipen, line 136: Fix TODO. Why ITE has the same "then" and "else" branches?

usvm-ts/src/main/kotlin/org/usvm/machine/operator/TsBinaryOperator.kt
```diff
@@ -196,7 +209,10 @@ sealed interface TsBinaryOperator {
                             )
 
                             // TODO: 'fake(ref)' == 'bool'
```
Lipen, line 211: Is this TODO implemented? Remove or move.

usvm-ts/src/main/kotlin/org/usvm/machine/operator/TsBinaryOperator.kt
```diff
@@ -196,7 +209,10 @@ sealed interface TsBinaryOperator {
...
-                            scope.assert(lhsType.refTypeExpr.not())
+                            conjuncts += ExprWithTypeConstraint(
+                                constraint = lhsType.refTypeExpr,
+                                expr = mkFalse() // TODO mistake, we should coerce the ref object
```
Lipen, line 214: Fix TODO

usvm-ts/src/main/kotlin/org/usvm/machine/operator/TsBinaryOperator.kt
```diff
@@ -488,46 +525,47 @@ sealed interface TsBinaryOperator {
...
-                    scope.assert(condition)
                 }
+                else -> error("Should not be called")
             }
 
             return with(Eq) {
-                resolveFakeObject(lhs, rhs, scope)
+                mkAnd(
+                    typeConstraint,
+                    resolveFakeObject(lhs, rhs, scope)
```
Lipen, line 567: It is better to extract the "loose equality" result into a separate variable and conjunct it with `typeConstraints` afterwards. This would better reflect the delegation semantics of StrictEq.

usvm-ts/src/test/kotlin/org/usvm/samples/FieldAccess.kt
```diff
@@ -77,8 +77,8 @@ class FieldAccess : TsMethodTestRunner() {
                 x.number == 1.1 && r.number == 14.0
             },
             { a, r ->
-                val x = a.properties["x"] as TsTestValue.TsNumber
-                x.number != 1.1 && r.number == 10.0
+                val x = a.properties["x"] as? TsTestValue.TsNumber
```
Lipen, line 80: Why "as?" cast? We expect `a.properties["x"]` to be `number` in all execution branches, so we cast it in both properties that we check. Why changing it to `as?` cast only in one?

usvm-ts/src/test/kotlin/org/usvm/util/Util.kt
```diff
@@ -0,0 +1,3 @@
+package org.usvm.util
+
+fun Boolean.toFp() = if (this) 1.0 else 0.0
```
Lipen, line 3: This is not "fp", this is `kotlin.Double`.

Lipen (changes requested):

usvm-ts/src/test/kotlin/org/usvm/samples/FieldAccess.kt
CaelmBleidd: Why? The source code: 
```
    conditionalFieldAccess(a: SimpleClass): number {
        if (a.x === 1.1) return 14;
        return 10;
    }

class SimpleClass {
    x: any = 5;
}

```

x has `any` type, for example, undefined can be chosen as a parameter for a test

Lipen (approved):
Seems good enough

PR #278, opened 2025-05-20
Different fixes

CaelmBleidd:
Changes list: 

1. Rename FakeType into EtsFakeType
2. Introduce TsVoidValue and TsVoidSort
3. Remove constraints for the exact subtypes, replace with structural type equality
4. Introduce EtsFieldResolutionResult
5. Create a fake field if the real one cannot be resolved
6. Add processing of missing local variables: they are represented as fields of some global object now
7. Mock calls of methods without entry points
8. Add error processing for if stmts without two successors
9. Support void value as a return type for mocked methods
10. Add implementation for sub operator 
11. Rename AuxiliaryType into EtsAuxiliaryType
12. Rewrite EtsTypeSystem
13. Support aliasType in the TypeToSort function
14. Rewrite EtsHierarchy
15. Add a new test for Photos project
16. Add new tests
17. Change type resolution in TsTestResolver


Lipen:
Please, provide some PR description, listing at least some changes, thank you!

CaelmBleidd:
> Please, provide some PR description, listing at least some changes, thank you!

Done @Lipen

PR #282, opened 2025-05-28
Replaced access path bases with access paths in backward dataflow

MForest7:
Replaced nested objects in backward dataflow with access paths, also get a performance increase.

|project|stmts|avg time (main)|avg time (new)|aa improvement, % (old)|aa improvement, % (new)|
|:--|:--|:--|:--|:--|:--|
|project1|9914|2.5s|1.6s|37.2|39.4|
|project2|26679|2.2s|1.4s|15.3|16.5|
|project3|4272|0.8s|0.4s|29.0|28.8|
|project4|4584|0.5s|0.3s|37.5|35.6|
|project5|121091|30.4s|9.3s|35.3|37.4|
|project6|28548|3.3s|1.7s|36.8|38.9|
|project7|5162|0.6s|0.4s|28.1|28.0|

### DIFF
--- a/usvm-ts-dataflow/src/test/resources/ts/testcases.ts
+++ b/usvm-ts-dataflow/src/test/resources/ts/testcases.ts
@@ -799,17 +799,17 @@ class CaseNew {
     }
 }
 
-class CaseNew2 {
-    entrypoint() {
-        let x = new Something(); // unresolved
-        console.log(x.foo);
-        this.infer(x);
-    }
-
-    infer(a: any): any {
-        const EXPECTED_ARG_0 = "Something { foo: any }";
-    }
-}
+// class CaseNew2 {
+//     entrypoint() {
+//         let x = new Something(); // unresolved
+//         console.log(x.foo);
+//         this.infer(x);
+//     }
+//
+//     infer(a: any): any {
+//         const EXPECTED_ARG_0 = "Something { foo: any }";
+//     }
+// }
 
 // ----------------------------------------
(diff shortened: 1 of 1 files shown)


Lipen:
What does the `%` mean in the table? Replace the table with fewer columns: project, size (stmts or similar), old time (avg +- sd), new time (avg +- sd). If the deviation is low (and it should be low, since the analysis should be near-deterministic), it can be omitted.

What about the "quality"? Comparing only the performance is not enough for such major change. Pick a metric (number of inferred types, ratio of all type, ratio of annotated type, number of inferred calls, etc) and add "old/new quality" columns.

Lipen:
For the future reference: this PR disrupted the "saved types" trick for handling `new T` with unresolved `T`, in particular, the filling of `typeInfo` (later passed to `savedTypes` in Forward FF) in Backward FF was removed since it is either can't be expressed anymore, or it is just not clear how to do it straight. Consequently, I have removed the whole "saved types" logic from Forward FF to clean-up the code.

Below is the failing test with unresolved `T` and "expected" type which could be passed only with saved types trick, but it is not clean whether we even need to support such case. So, we have decided to currently disable such test.
```ts
class CaseNew2 {
    entrypoint() {
        let x = new Something(); // unresolved
        let y = x.foo; // use 'foo' field
        this.infer(x);
    }
    infer(a: any): any {
        const EXPECTED_ARG_0 = "Something { foo: any }";
    }
}
```

PR #285, opened 2025-06-03
Input arrays support

CaelmBleidd:
The following changes were made: 
1. New matchers for tests supporting executions number checker
2. Support for arrays as parameters
3. Register fake objects creation to support their restoration during model resolving 
4. Reading by negative indices and indices out of range are prohibited. We'll treat this operation as an error

### DIFF
--- a/usvm-ts/src/main/kotlin/org/usvm/machine/expr/TsExprResolver.kt
+++ b/usvm-ts/src/main/kotlin/org/usvm/machine/expr/TsExprResolver.kt
@@ -788,7 +788,7 @@ class TsExprResolver(
                 val fakeType = instanceRef.getFakeType(scope)
 
                 // If we want to get length from a fake object, we assume that it is an array.
-                scope.assert(fakeType.refTypeExpr)
+                scope.doWithState { pathConstraints += fakeType.refTypeExpr }
 
                 val refLValue = getIntermediateRefLValue(instanceRef.address)
                 val obj = scope.calcOnState { memory.read(refLValue) }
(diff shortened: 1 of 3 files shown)


usvm-ts/src/main/kotlin/org/usvm/machine/expr/TsExprResolver.kt
```diff
@@ -696,32 +774,46 @@ class TsExprResolver(
...
-                    val length = scope.calcOnState { memory.read(lengthLValue) }
-                    return mkBvToFpExpr(
-                        fp64Sort,
-                        fpRoundingModeSortDefaultValue(),
-                        length.asExpr(sizeSort),
-                        signed = true
-                    )
+
+                // If we want to get length from a fake object, we assume that it is an array.
+                scope.assert(fakeType.refTypeExpr)
```
Lipen, line 790: Why `assert` and not `pathConstraints += ...`?

usvm-ts/src/main/kotlin/org/usvm/machine/expr/TsExprResolver.kt
```diff
@@ -809,7 +901,12 @@ class TsExprResolver(
                 blockOnFalseState = allocateException(EtsStringType) // TODO incorrect exception type
             )
 
-            val arrayType = EtsArrayType(EtsUnknownType, 1) // TODO: expr.type
+            val arrayType = expr.type as EtsArrayType
```
Lipen, line 904: This cast might fail.

PR #296, opened 2025-06-24
Bump Kotlin to 2.2.0

Lipen:
This PR updates Kotlin version to `2.2.0`.

Due to some internal changes of Kotlin compiler and its type inference, it required to explicitly specify some generics on usage side.


CaelmBleidd:
Right now it is not necessary a change for USVM

PR #302, opened 2025-07-08
[TS] Less expression 

CaelmBleidd:
`<` and `>` operators extended support

### DIFF
--- a/usvm-ts/src/test/kotlin/org/usvm/samples/operators/Add.kt
+++ b/usvm-ts/src/test/kotlin/org/usvm/samples/operators/Add.kt
@@ -1,6 +1,7 @@
 package org.usvm.samples.operators
 
 import org.jacodb.ets.model.EtsScene
+import org.junit.jupiter.api.Disabled
 import org.usvm.api.TsTestValue
 import org.usvm.util.TsMethodTestRunner
 import org.usvm.util.toDouble
@@ -75,6 +76,7 @@ class Add : TsMethodTestRunner() {
         )
     }
 
+    @Disabled("Flaky test, see https://github.com/UnitTestBot/usvm/issues/310")
     @Test
     fun `add unknown values`() {
         val method = getMethod(className, "addUnknownValues")


usvm-ts/src/test/resources/samples/operators/Less.ts
```diff
@@ -0,0 +1,63 @@
...
+        return 0; // Return 0 if they are equal
+    }
+
+    lessRefs(a: object, b: object): boolean {
+        if (a < b) {
+            return true; // Reference comparison, not value
+        }
+
+        if (b < a) {
+            return false; // Reference comparison, not value
```
Lipen, line 44: What is the point of this return? Consider splitting the test samples to make test actually "test one thing at a time".

usvm-ts/src/main/kotlin/org/usvm/machine/operator/TsBinaryOperator.kt
```diff
@@ -97,294 +99,274 @@ sealed interface TsBinaryOperator {
...
 
-    data object Eq : TsBinaryOperator {
-        override fun TsContext.onBool(
-            lhs: UBoolExpr,
-            rhs: UBoolExpr,
-            scope: TsStepScope,
-        ): UBoolExpr {
-            return mkEq(lhs, rhs)
-        }
+    fun <R : USort> TsBinaryOperator.commonResolveFakeObject(
```
Lipen, line 102: This function does not need to be an extension of `TsBinaryOperator`, since this is already a method of this class. Weird how IDEA does not warn about this meaningless construction.

You probably want to make it an extension of `TsContext`, similar to `onXXX` methods defined above. That way, you could eliminate nasty `with (lhs.tctx)` block.

usvm-ts/src/main/kotlin/org/usvm/machine/operator/TsBinaryOperator.kt
```diff
@@ -97,294 +99,274 @@ sealed interface TsBinaryOperator {
...
+                        rhs.extractBool(scope),
+                        scope
+                    )?.asExpr(bottomValueSort) ?: return null
+                )
 
-            val conjuncts = mutableListOf<ExprWithTypeConstraint<UBoolSort>>()
+                // fake(fp) + fake(fp)
+                conjuncts += ExprWithTypeConstraint(
+                    constraint = mkAnd(lhsType.fpTypeExpr, rhsType.fpTypeExpr),
+                    expr = onFp(lhs.extractFp(scope), rhs.extractFp(scope), scope)?.asExpr(bottomValueSort) ?: return null
```
Lipen, line 153: Consider placing arguments on separate lines, similar to above. Avoid long lines and inconsistent code.

usvm-ts/src/main/kotlin/org/usvm/machine/types/EtsFakeType.kt
```diff
@@ -43,7 +43,7 @@ class EtsFakeType(
     }
 }
 
-data class ExprWithTypeConstraint<Sort : USort>(
+data class ExprWithTypeConstraint<T : USort>(
```
Lipen, line 46: This change is not necessary.

PR #314, opened 2025-07-27
Updated approximations

MchKosticyn:


usvm-jvm/src/main/kotlin/org/usvm/machine/state/JcStateUtils.kt
```diff
@@ -88,3 +92,29 @@ fun JcState.skipMethodInvocationWithValue(methodCall: JcMethodCall, value: UExpr
...
+fun JcState.skipMethodInvocationAndBoxIfNeeded(methodCall: JcMethodCall, valueType: JcType, value: UExpr<out USort>) {
+    val typeSystem = ctx.typeSystem<JcType>()
+    val methodReturnType = ctx.cp.findType(methodCall.method.returnType.typeName)
+    when {
+        valueType is JcPrimitiveType && methodReturnType is JcClassType -> {
+            val boxedType = valueType.autoboxIfNeeded() as JcClassType
+            check(typeSystem.isSupertype(methodReturnType, boxedType)) {
+                "skipMethodInvocationAndBoxIfNeeded: Incorrect method return type"
+            }
+            val boxMethod = boxedType.declaredMethods.find {
```
Saloed, line 105: Please, use first instead of find to avoid !!

usvm-jvm/src/main/kotlin/org/usvm/machine/JcApproximations.kt
```diff
@@ -956,6 +1033,15 @@ class JcMethodApproximationResolver(
                 val arg = it.arguments.single().asExpr(ctx.booleanSort)
                 scope.assert(arg)?.let { ctx.voidValue }
             }
+            dispatchUsvmApiMethod(Engine::assumeSymbolic) {
```
Saloed, line 1036: Seems that we should update soft constraints here instead of assert

Saloed (changes requested):

Saloed (approved):

PR #316, opened 2025-07-30
`initializeAllocatedSet` in `USetRegion`

tochilinak:


usvm-core/src/main/kotlin/org/usvm/collection/set/primitive/USymbolicSetId.kt
```diff
@@ -115,6 +119,40 @@ class UAllocatedSetId<SetType, ElementSort : USort, Reg : Region<Reg>>(
     }
 
     override fun hashCode(): Int = hash(setAddress, setType, elementSort)
+
+    fun initializedSet(
+        content: Set<UExpr<ElementSort>>,
+        guard: UBoolExpr,
+        makeDisjointCheck: Boolean = true,
```
Saloed, line 126: Do we really need this flag? As for me, it is better to always check elements disjointness , or remove this check at all and add comment about elements.

Saloed (changes requested):
Please, rebase

usvm-core/src/main/kotlin/org/usvm/collection/set/primitive/USymbolicSetId.kt
tochilinak: In case of `SetRegion`, `unite` is linear --- because ordinary set is used instead of persistent one. So this check leads to quadratic complexity, which is too slow. This check is still usefull for debugging though.

For other regions, this check can be fine.

tochilinak: Done

Saloed (approved):

PR #317, opened 2025-07-31
Use jacodb from GitHub Packages Registry

Lipen:
This PR switches from using JitPack for jacodb artifacts to using GitHub Package Registry.

**NOTE:** Using GPR as a consumer (that is, download JARs from repository) requires an authorization via a GitHub token with `packages: read` permissions.  The local setup is very simple:
- Generate a new _**classic**_ token at ["GitHub > Settings > Developer Settings > Personal access tokens > Tokens (classic)"](https://github.com/settings/tokens). Tick only the "`read:packages`" permissions.
- Set `gpr.user` and `gpr.key` Gradle properties globally in `~/.gradle/gradle.properties` file. For example:
  ```
  gpr.user=your-github-username
  gpr-key=your-github-token
  ```
- Compile the project. All deps from GPR should resolve successfully.


buildSrc/src/main/kotlin/usvm.kotlin-conventions.gradle.kts
```diff
@@ -14,6 +14,8 @@ repositories {
     maven("https://jitpack.io")
 }
 
+setupJacodbGitHubPackagesRepository()
```
Saloed, line 17: Do we really need jacodb repo in all projects?

Saloed (changes requested):
Please, review jvm test issues

Lipen:
> Please, review jvm test issues

I can reproduce it locally. Also, if I build jacodb to the local maven repo using JDK 8, and then use those artifacts in USVM (running with jdk 11, since usvm requires it), tests pass. Thus, it is the issue related to the built artifacts. I already have heard the rumors that such situations might happen, and that "compiling with java 8 compatibility" is no the same as "compiling with JDK 8 directly", but never seen this in reality.

@Saloed What is the action here? Again just hide the problem by using jdk 8 for building jacodb on CI?

Lipen:
He-he, similar issue: https://github.com/javalin/javalin/issues/2117

PR #323, opened 2025-08-19
[TS] Support globals and imports

Lipen:
This PR adds support for global variables and imported symbols.

### DIFF
--- a/usvm-ts/src/main/kotlin/org/usvm/machine/types/FakeExprUtil.kt
+++ b/usvm-ts/src/main/kotlin/org/usvm/machine/types/FakeExprUtil.kt
@@ -16,6 +16,7 @@ import org.usvm.machine.state.TsState
 import org.usvm.memory.ULValue
 
 fun TsState.mkFakeValue(
+    scope: TsStepScope?, // pass `null` only in the initial state, where `scope` is not available!
     boolValue: UBoolExpr? = null,
     fpValue: UExpr<KFp64Sort>? = null,
     refValue: UHeapRef? = null,
@@ -43,7 +44,12 @@ fun TsState.mkFakeValue(
         refTypeExpr = refTypeExpr,
     )
     memory.types.allocate(address, type)
-    pathConstraints += type.mkExactlyOneTypeConstraint(ctx)
+    val constraint = type.mkExactlyOneTypeConstraint(ctx)
+    if (scope != null) {
+        scope.assert(constraint)
+    } else {
+        pathConstraints += constraint
+    }
 
     if (boolValue != null) {
         val boolLValue = ctx.getIntermediateBoolLValue(address)
--- a/usvm-ts/src/main/kotlin/org/usvm/api/TsMock.kt
+++ b/usvm-ts/src/main/kotlin/org/usvm/api/TsMock.kt
@@ -24,9 +24,10 @@ fun mockMethodCall(
 
                 is TsUnresolvedSort -> scope.calcOnState {
                     mkFakeValue(
-                        makeSymbolicPrimitive(ctx.boolSort),
-                        makeSymbolicPrimitive(ctx.fp64Sort),
-                        makeSymbolicRefUntyped()
+                        scope = scope,
+                        boolValue = makeSymbolicPrimitive(ctx.boolSort),
+                        fpValue = makeSymbolicPrimitive(ctx.fp64Sort),
+                        refValue = makeSymbolicRefUntyped(),
                     )
                 }
 
--- a/usvm-ts/src/main/kotlin/org/usvm/machine/expr/ReadArray.kt
+++ b/usvm-ts/src/main/kotlin/org/usvm/machine/expr/ReadArray.kt
@@ -126,7 +126,7 @@ fun TsContext.readArray(
         if (ref.isFakeObject()) {
             ref
         } else {
-            val fakeObj = mkFakeValue(bool, fp, ref)
+            val fakeObj = mkFakeValue(scope, bool, fp, ref)
             lValuesToAllocatedFakeObjects += refLValue to fakeObj
             memory.write(refLValue, fakeObj, guard = trueExpr)
             fakeObj
--- a/usvm-ts/src/main/kotlin/org/usvm/machine/expr/ReadField.kt
+++ b/usvm-ts/src/main/kotlin/org/usvm/machine/expr/ReadField.kt
@@ -111,7 +111,7 @@ internal fun TsContext.readField(
         if (ref.isFakeObject()) {
             ref
         } else {
-            val fakeObj = mkFakeValue(bool, fp, ref)
+            val fakeObj = mkFakeValue(scope, bool, fp, ref)
             lValuesToAllocatedFakeObjects += refLValue to fakeObj
             memory.write(refLValue, fakeObj, guard = trueExpr)
             fakeObj
--- a/usvm-ts/src/main/kotlin/org/usvm/machine/interpreter/TsInterpreter.kt
+++ b/usvm-ts/src/main/kotlin/org/usvm/machine/interpreter/TsInterpreter.kt
@@ -774,7 +774,7 @@ class TsInterpreter(
                 val bool = mkRegisterReading(idx, boolSort)
                 val fp = mkRegisterReading(idx, fp64Sort)
                 val ref = mkRegisterReading(idx, addressSort)
-                val fakeObject = state.mkFakeValue(bool, fp, ref)
+                val fakeObject = state.mkFakeValue(null, bool, fp, ref)
                 val lValue = mkRegisterStackLValue(addressSort, idx)
                 state.memory.write(lValue, fakeObject.asExpr(addressSort), guard = trueExpr)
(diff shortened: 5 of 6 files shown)


usvm-ts/src/main/kotlin/org/usvm/machine/expr/TsLocalResolver.kt
```diff
@@ -0,0 +1,199 @@
...
+}
+
+fun TsContext.resolveLocalToLValue(
+    scope: TsStepScope,
+    local: EtsValue,
+): ULValue<*, *>? {
+    val currentMethod = scope.calcOnState { lastEnteredMethod }
+
+    if (currentMethod.name == DEFAULT_ARK_METHOD_NAME) {
+        // TODO
```
CaelmBleidd, line 71: Add at least a comment, what to do here, and what this section means

usvm-ts/src/main/kotlin/org/usvm/machine/expr/TsLocalResolver.kt
```diff
@@ -0,0 +1,199 @@
...
+                // Try to get the saved sort for this imported dflt object field
+                val symbolNameInImportedFile = resolutionResult.exportInfo.originalName
+                val savedSort = scope.calcOnState {
+                    getSortForDfltObjectField(importedFile, symbolNameInImportedFile)
+                }
+
+                if (savedSort == null) {
+                    // No saved sort means this field was never assigned to, which is an error
+                    logger.error { "Trying to read unassigned imported symbol: '$local' from '$importedFile'" }
+                    scope.assert(falseExpr)
```
CaelmBleidd, line 175: If this is an error, you should throw an error. Not simply remove the state that will be an issue for someone later

usvm-ts/src/main/kotlin/org/usvm/machine/expr/TsLocalResolver.kt
```diff
@@ -0,0 +1,199 @@
...
+
+        // Try to get the saved sort for this dflt object field
+        val savedSort = scope.calcOnState {
+            getSortForDfltObjectField(file, local.name)
+        }
+
+        if (savedSort == null) {
+            // No saved sort means this field was never assigned to, which is an error
+            logger.error { "Trying to read unassigned global variable: '$local' in $file" }
+            scope.assert(falseExpr)
```
CaelmBleidd, line 131: If this is an error, you should throw an error. Not simply remove the state that will be an issue for someone later

usvm-ts/src/main/kotlin/org/usvm/machine/expr/TsLocalResolver.kt
```diff
@@ -0,0 +1,199 @@
...
+                    scope.assert(falseExpr)
+                    return null
+                }
+
+                mkFieldLValue(savedSort, importedDfltObject, symbolNameInImportedFile)
+            }
+
+            is SymbolResolutionResult.FileNotFound -> {
+                logger.error { "Cannot resolve import for '$local': ${resolutionResult.reason}" }
+                scope.assert(falseExpr)
```
CaelmBleidd, line 184: You should forget about this `assert(false)` pattern

usvm-ts/src/main/kotlin/org/usvm/machine/interpreter/TsGlobals.kt
```diff
@@ -0,0 +1,56 @@
...
+
+private fun TsContext.mkGlobalsInitializedFlag(
+    instance: UHeapRef,
+): UFieldLValue<String, UBoolSort> {
+    return mkFieldLValue(boolSort, instance, "__initialized__")
+}
+
+internal fun TsState.initializeGlobals(file: EtsFile) {
+    markGlobalsInitialized(file)
+    val dfltClass = file.classes.first { it.name == DEFAULT_ARK_CLASS_NAME }
```
CaelmBleidd, line 48: Why first?

usvm-ts/src/main/kotlin/org/usvm/machine/types/FakeExprUtil.kt
```diff
@@ -15,55 +15,52 @@ import org.usvm.machine.interpreter.TsStepScope
...
+        .takeIf { boolValue == null && fpValue == null && refValue != null }
+        ?: makeSymbolicPrimitive(boolSort)
+
+    val type = EtsFakeType(
+        boolTypeExpr = boolTypeExpr,
+        fpTypeExpr = fpTypeExpr,
+        refTypeExpr = refTypeExpr,
+    )
+    memory.types.allocate(address, type)
+    pathConstraints += type.mkExactlyOneTypeConstraint(ctx)
```
CaelmBleidd, line 46: This is not a simple change. Because of it later you can have inconsistent model somewhere. 

You have states: a -> b -> c, and in a you added this path constraint. But it doesn't recalculate models, therefore, when in C you asks info about this type constraint, it might be incorrect since you didn't call solver. Previously, there was an assert that forces this models reevaluation

usvm-ts/src/main/kotlin/org/usvm/util/TsImports.kt
```diff
@@ -0,0 +1,203 @@
...
+fun EtsScene.resolveImportInfo(
+    currentFile: EtsFile,
+    importInfo: EtsImportInfo,
+): SymbolResolutionResult {
+    return resolveSymbol(currentFile, importInfo.from, importInfo.originalName, importInfo.type)
+}
+
+private fun EtsScene.resolveSystemLibrary(importPath: String): ImportResolutionResult {
+    val foundFile = sdkFiles.find { file ->
+        file.signature.fileName.equals(importPath, ignoreCase = true) ||
```
CaelmBleidd, line 57: with (field.signature.filename)

CaelmBleidd (changes requested):

usvm-ts/src/main/kotlin/org/usvm/machine/expr/TsLocalResolver.kt
Lipen: WIP

usvm-ts/src/main/kotlin/org/usvm/machine/expr/TsLocalResolver.kt
Lipen: Unresolved imports are not a hard error. Specifically, on the current state of the development. Some files are missing even in the full sdk, and it is yet unclear what to do in such cases. But throwing a hard error and not producing any execution resulting is a blocker, I wont get any info about coverage in such case. Recall that currently we are still collecting the stats about what we support and we don't by running on real projects, so `throw IllegalStateException("sorry, your project or/and IR is garbage, fix it and come again for any results")` is a no-go solution.

usvm-ts/src/main/kotlin/org/usvm/machine/types/FakeExprUtil.kt
Lipen: That was your suggestion. How to re-organize all this without an access to the scope?

usvm-ts/src/main/kotlin/org/usvm/util/TsImports.kt
Lipen: Do not over-complicate things with those context functions...

usvm-ts/src/main/kotlin/org/usvm/machine/expr/TsLocalResolver.kt
Lipen: Done. It was required to adopt the reading of "locals" in `%dflt::%dflt`, similar to how we ad-hoc the assignments to those "locals": if the "local" variable _looks like a global_, then we should treat it as global, not local. In the perfect world, `method.body.locals` would contain the correct list of "real locals", and this ad-hoc would not be needed, but currently the IR is broken (see issue https://gitcode.com/openharmony-sig/arkanalyzer/issues/768).

For now, the following heuristic works for determining whether the variable "looks like a global". The only "real local" variable I have seen are: "temp locals" (`%0` from AA, and `_tmp0` from our 3AC builder) and the local named "this". All other variables in `%dflt` are considered global, both when reading AND writing them.

usvm-ts/src/main/kotlin/org/usvm/machine/state/TsStateUtils.kt
```diff
@@ -33,3 +32,16 @@ inline val EtsMethod.parametersWithThisCount: Int
...
 inline val EtsMethod.localsCount: Int
     get() = locals.size
+
+// TODO: fix handling of arguments and use this function in machine
+fun TsState.makeCall(
+    method: EtsMethod,
+    instance: UExpr<*>,
+    args: List<UExpr<*>>,
+) {
+    pushSortsForArguments(0) { null }
```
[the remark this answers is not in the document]
Lipen: This utility function is not quite correct, so I have not transitioned the code to use it instead of inlining these several lines each time you want to make a call. So, currently, this function is unused, and since I cannot come up with the correct solution for it right now, I have just removed it.

PR #327, opened 2025-09-05
[TS] Fixes with fake objects inside `UIteExpr`

CaelmBleidd:


Lipen:
Please, add unit tests that fail with the current `main`, and which are "solved" by this PR. If the tests could not be implemented, at least mention those situations in the _PR description_.

PR #336, opened 2025-09-12
[TS] Fixes in TS interpreter for real projects

Lipen:
This PR introduces some minor fixes to make USVM TS machine work slightly more smoothly on the real project.

### DIFF
--- a/usvm-ts/src/main/kotlin/org/usvm/machine/expr/ExprUtil.kt
+++ b/usvm-ts/src/main/kotlin/org/usvm/machine/expr/ExprUtil.kt
@@ -249,7 +249,7 @@ fun TsContext.checkReadingInRange(
     )
 }
 
-fun TsContext.checkLengthBounds(
+fun TsContext.ensureLengthBounds(
     scope: TsStepScope,
     length: UExpr<TsSizeSort>,
     maxLength: Int,
--- a/usvm-ts/src/main/kotlin/org/usvm/machine/expr/ReadArray.kt
+++ b/usvm-ts/src/main/kotlin/org/usvm/machine/expr/ReadArray.kt
@@ -1,6 +1,7 @@
 package org.usvm.machine.expr
 
 import io.ksmt.utils.asExpr
+import mu.KotlinLogging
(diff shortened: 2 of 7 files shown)


usvm-ts/src/main/kotlin/org/usvm/machine/expr/ExprUtil.kt
```diff
@@ -248,3 +248,16 @@ fun TsContext.checkReadingInRange(
...
+    scope: TsStepScope,
+    length: UExpr<TsSizeSort>,
+    maxLength: Int,
+): Unit? {
+    // Check that length is non-negative and does not exceed `maxLength`.
+    val condition = mkAnd(
+        mkBvSignedGreaterOrEqualExpr(length, mkBv(0)),
+        mkBvSignedLessOrEqualExpr(length, mkBv(maxLength))
+    )
+    return scope.assert(condition)
```
CaelmBleidd: It is different conditions. Reading by a negative element should raise an exception, since it is probably an error. Reading by 1001 is not an error in all cases

CaelmBleidd (changes requested):

usvm-ts/src/main/kotlin/org/usvm/machine/expr/ExprUtil.kt
Lipen: I remember, we decided that reading out of bounds is equally wrong in both cases. If we had decided to allow reading indices larger than the length, then it would have already been implemented, as it would simply return undefined.

~~I do agree, however, that it should use `scope.fork` and throw an exception, instead of just asserting.~~

Lipen: Note: there are two different functions above this, which are intended for array reading: `checkNegativeIndexRead` and `checkReadingInRange`. They fork and throw on out-of-bounds access.
The function `ensureLengthBounds` that you have commented in, is for another case: when we read a symbolic length of an array, we explicitly force this symbolic value to be in range 0..length. Forking and throwing exception here does not make sense.

CaelmBleidd (approved):

PR #343, opened 2026-07-29
Less restrictions in the memory

metametamoon:
Relaxed the modifiers (added open, downgraded access modifiers) so that it is possible to create a memory-local addresses allocations (which allows for path selector-independent addresses, helpful for debugging).


tochilinak (approved):

metametamoon:
Does not look as easy as I thought to implement in TVM due to the specifics of memory implementation.

PR #344, opened 2026-08-21
Adopt JacoDB native TypeScript frontend in USVM

CaelmBleidd:
## Summary

- Make the merged JacoDB native TypeScript frontend the default frontend for `.ts` sources and pin the published JacoDB `neo` artifact `9ea33879c9`.
- Add opt-in local composite substitution through `-PuseLocalJacodb[=<path>]`, so JacoDB frontend changes can be tested before a new remote artifact is available.
- Stabilize the legacy ArkAnalyzer resource-generation bootstrap still present in `ci-ts` by pinning `@types/node@18` before installation; its bundled TypeScript 4.9 cannot parse the latest floating Node declarations.
- Adapt symbolic execution and TypeScript dataflow to the native frontend's IR:
  - map literal types to their runtime base sorts and advance through explicit NOP nodes;
  - execute pre/post increment and decrement expressions;
  - canonicalize array descriptors, bind closure environments, initialize native module fields, and preserve import/field behavior;
  - resolve native pointer calls for the built-in `Number`, including `Number()` returning `0`;
  - preserve nominal `instanceof` alternatives and run the type guesser to a fixed point;
  - execute overload implementations in USVM, and canonicalize static overload groups before forking so bodyless declarations do not create equivalent states;
  - support concrete string concatenation, including ECMAScript-compatible formatting of finite fractional and extreme `number` values.
- Add focused regression tests for the migrated IR shapes and execution behavior, including overload state counts and concrete number formatting boundaries.

## Architectural boundary

JacoDB preserves TypeScript declarations and structural types. It does not choose which overload is executable. Runtime overload selection and state-forking policy remain in USVM.

## Verification

The final integration test uses the JacoDB artifact published by JitPack, with no local composite substitution:

```bash
./gradlew \
  :usvm-ts:test \
  :usvm-ts-dataflow:test
```

Result: `BUILD SUCCESSFUL` (28 actionable tasks). This verifies both remote dependency resolution and the native frontend integration.

Detekt was also run under JDK 21 while developing the migration:

```bash
env JAVA_HOME=/path/to/jdk-21 \
./gradlew \
  -PuseLocalJacodb=/path/to/jacodb \
  :usvm-ts:detekt
```

Result: `BUILD SUCCESSFUL`; no new warnings remain in the overload or number-formatting changes.

The CI ArkAnalyzer bootstrap was reproduced from a clean `neo/2025-09-03` clone with the same manifest pin; `npm install` and `npm run build` both completed successfully.

JacoDB PR UnitTestBot/jacodb#363 is merged into `neo`. JitPack reports merge commit `9ea33879c9` as successfully published, and its `jacodb-ets` JAR contains the bundled TypeScript frontend runtime.

### DIFF
--- a/usvm-ts/src/main/kotlin/org/usvm/machine/expr/CallStaticApproximations.kt
+++ b/usvm-ts/src/main/kotlin/org/usvm/machine/expr/CallStaticApproximations.kt
@@ -35,6 +35,9 @@ private fun TsExprResolver.handleR(): UExpr<*> = with(ctx) {
 }
 
 internal fun TsExprResolver.handleNumberConverter(args: List<EtsValue>): UExpr<*>? = with(ctx) {
+    if (args.isEmpty()) {
+        return mkFp64(0.0)
+    }
     check(args.size == 1) {
         "Number() should have exactly one argument, but got ${args.size}"
     }
--- a/usvm-ts/src/main/kotlin/org/usvm/machine/expr/CallStaticApproximations.kt
+++ b/usvm-ts/src/main/kotlin/org/usvm/machine/expr/CallStaticApproximations.kt
@@ -1,6 +1,7 @@
 package org.usvm.machine.expr
 
 import org.jacodb.ets.model.EtsStaticCallExpr
+import org.jacodb.ets.model.EtsValue
 import org.usvm.UExpr
 import org.usvm.machine.expr.TsExprApproximationResult.Companion.from
 
@@ -14,7 +15,7 @@ internal fun TsExprResolver.tryApproximateStaticCall(
 
     // Handle `Number(...)` calls
     if (expr.callee.name == "Number") {
-        return from(handleNumberConverter(expr))
+        return from(handleNumberConverter(expr.args))
     }
 
     // Handle `Boolean(...)` calls
@@ -33,11 +34,11 @@ private fun TsExprResolver.handleR(): UExpr<*> = with(ctx) {
     mockSymbol
 }
 
-private fun TsExprResolver.handleNumberConverter(expr: EtsStaticCallExpr): UExpr<*>? = with(ctx) {
-    check(expr.args.size == 1) {
-        "Number() should have exactly one argument, but got ${expr.args.size}"
+internal fun TsExprResolver.handleNumberConverter(args: List<EtsValue>): UExpr<*>? = with(ctx) {
+    check(args.size == 1) {
+        "Number() should have exactly one argument, but got ${args.size}"
     }
-    val arg = resolve(expr.args.single()) ?: return null
+    val arg = resolve(args.single()) ?: return null
     return mkNumericExpr(arg, scope)
 }
 
--- a/usvm-ts/src/main/kotlin/org/usvm/machine/expr/TsExprResolver.kt
+++ b/usvm-ts/src/main/kotlin/org/usvm/machine/expr/TsExprResolver.kt
@@ -249,23 +249,44 @@ class TsExprResolver(
     }
 
     override fun visit(expr: EtsPostIncExpr): UExpr<out USort>? {
-        logger.warn { "visit(${expr::class.simpleName}) is not implemented yet" }
-        error("Not supported $expr")
+        return resolveUpdateExpression(expr.arg, increment = true, returnOldValue = true)
     }
 
     override fun visit(expr: EtsPostDecExpr): UExpr<out USort>? {
-        logger.warn { "visit(${expr::class.simpleName}) is not implemented yet" }
-        error("Not supported $expr")
+        return resolveUpdateExpression(expr.arg, increment = false, returnOldValue = true)
     }
 
     override fun visit(expr: EtsPreIncExpr): UExpr<out USort>? {
-        logger.warn { "visit(${expr::class.simpleName}) is not implemented yet" }
-        error("Not supported $expr")
+        return resolveUpdateExpression(expr.arg, increment = true, returnOldValue = false)
     }
 
     override fun visit(expr: EtsPreDecExpr): UExpr<out USort>? {
-        logger.warn { "visit(${expr::class.simpleName}) is not implemented yet" }
-        error("Not supported $expr")
+        return resolveUpdateExpression(expr.arg, increment = false, returnOldValue = false)
+    }
+
+    private fun resolveUpdateExpression(
+        target: EtsEntity,
+        increment: Boolean,
+        returnOldValue: Boolean,
+    ): UExpr<out USort>? = with(ctx) {
+        val oldValue = resolve(target) ?: return null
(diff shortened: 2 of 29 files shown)


usvm-ts/src/main/kotlin/org/usvm/machine/expr/CallStaticApproximations.kt
```diff
@@ -33,11 +34,14 @@ private fun TsExprResolver.handleR(): UExpr<*> = with(ctx) {
     mockSymbol
 }
 
-private fun TsExprResolver.handleNumberConverter(expr: EtsStaticCallExpr): UExpr<*>? = with(ctx) {
-    check(expr.args.size == 1) {
-        "Number() should have exactly one argument, but got ${expr.args.size}"
+internal fun TsExprResolver.handleNumberConverter(args: List<EtsValue>): UExpr<*>? = with(ctx) {
+    if (args.isEmpty()) {
+        return mkFp64(0.0)
     }
```
CaelmBleidd, line 40: Why? Is it a real semantics?

usvm-ts/src/main/kotlin/org/usvm/machine/expr/TsExprResolver.kt
```diff
@@ -249,23 +254,44 @@ class TsExprResolver(
...
     override fun visit(expr: EtsPostIncExpr): UExpr<out USort>? {
-        logger.warn { "visit(${expr::class.simpleName}) is not implemented yet" }
-        error("Not supported $expr")
+        return resolveUpdateExpression(expr.arg, increment = true, returnOldValue = true)
     }
 
     override fun visit(expr: EtsPostDecExpr): UExpr<out USort>? {
-        logger.warn { "visit(${expr::class.simpleName}) is not implemented yet" }
-        error("Not supported $expr")
+        return resolveUpdateExpression(expr.arg, increment = false, returnOldValue = true)
```
CaelmBleidd: Let's use ENUM or sealed class instead of `increment = true`

usvm-ts/src/main/kotlin/org/usvm/machine/expr/TsExprResolver.kt
```diff
@@ -543,9 +569,30 @@ class TsExprResolver(
...
+        if (expr.type == EtsStringType) {
+            return resolveAfterResolved(expr.left, expr.right) { lhs, rhs ->
+                val lhsString = concreteStringValue(lhs)
+                    ?: error("Symbolic string concatenation is not supported for left operand: $lhs")
+                val rhsString = concreteStringValue(rhs)
+                    ?: error("Symbolic string concatenation is not supported for right operand: $rhs")
+                ctx.mkStringConstant(lhsString + rhsString, scope)
+            }
+        }
         return resolveBinaryOperator(TsBinaryOperator.Add, expr)
```
CaelmBleidd, line 586: Previously concatenation for strings was not supported at all?

usvm-ts/src/main/kotlin/org/usvm/machine/expr/TsExprResolver.kt
```diff
@@ -1179,7 +1252,16 @@ class TsSimpleValueResolver(
                 return null
             }
             val method = methods.single()
-            val ref = scope.calcOnState { getMethodRef(method) }
+            val closureParameter = method.parameters
+                .firstOrNull()
+                ?.takeIf { it.type is EtsLexicalEnvType }
+            val closure: UHeapRef? = if (closureParameter != null) {
+                val resolved = resolveLocal(EtsLocal(closureParameter.name, closureParameter.type)) ?: return null
```
CaelmBleidd: Let's extract creation of objects from expressions

usvm-ts/src/main/kotlin/org/usvm/machine/types/TsTypeSystem.kt
```diff
@@ -182,16 +182,10 @@ class TsTypeSystem(
...
-
-            return superClasses.any { cls ->
-                val properties = cls.getAllPropertiesCombined(hierarchy)
-
-                return properties.containsAll(unwrappedType.properties)
-            }
+            // An auxiliary type denotes a structural requirement, not a concrete
+            // nominal class. A class can satisfy that requirement (handled above),
+            // but the requirement itself is not a subtype of that class.
+            return false
```
CaelmBleidd: What? I don't get it. The idea of this class is that it is a subclass of any classes with the same set of fields since { a : Number } can be passed in a methods instead of { a: Number, b: Number }, etc. So I think that this is odd implementation and should be reverted

usvm-ts/src/main/kotlin/org/usvm/machine/expr/TsExprResolver.kt
CaelmBleidd: Replaced the boolean with UpdateOperator.INCREMENT and UpdateOperator.DECREMENT in 912de523.

usvm-ts/src/main/kotlin/org/usvm/machine/types/TsTypeSystem.kt
CaelmBleidd: Reverted the change and restored the established auxiliary-type relation. The original regression came from using structural assignability for runtime instanceof: the constraint engine then treated a nominal check as redundant. Commit 912de523 introduces an internal EtsNominalType only for runtime hierarchy checks, so structural assignability remains unchanged while Child and Parent alternatives are preserved. Added a regression test for the auxiliary relation.

usvm-ts/src/main/kotlin/org/usvm/machine/expr/TsExprResolver.kt
CaelmBleidd: Extracted the EtsLocal construction into the named closureLocal variable before resolving it in 912de523.

usvm-ts/src/main/kotlin/org/usvm/machine/expr/TsExprResolver.kt
CaelmBleidd: Yes. Before this branch, TsBinaryOperator.Add explicitly left string concatenation unsupported and followed numeric coercion paths. This change supports only fully concrete string concatenation; symbolic string operands remain unsupported.

usvm-ts/src/main/kotlin/org/usvm/machine/expr/CallStaticApproximations.kt
CaelmBleidd: Yes. This is the ECMAScript Number() semantics: when the argument is absent, the result is positive zero. See https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-number-constructor-number-value. Numeric.emptyNumber covers this case.

PR #377, opened 2026-08-28
[TS Calls] Add guarded semantic models with a simplified contract

CaelmBleidd:
### Capability work for #365

- add guarded semantic-model selection and execution to the existing unknown-call boundary
- preserve normal and exceptional completion, deferred state changes, aliases, residual fallback, and one observer decision per application
- add `Array.shift` as the intrinsic example backed by symbolic-memory `memcpy`
- require stable call-target evidence and a supported receiver representation; unknown and fake receivers use fallback

### Contract simplification gate for #383

Starting from `fad3d155` (the rebased last pre-simplification head after the original #365 registry and intrinsic packaging), this PR:

- collapses descriptor, registration, backend, and implementation wrappers into one `TsUnknownCallModel` object with an ID, declarative target, and guarded execution
- removes implementation-kind tags, supported-domain IDs, and stored precision values; residual presence expresses partial handling
- replaces the old profile and registry layers with one immutable enabled-model catalog and one residual fallback setting
- derives observable outcomes from model or residual decisions instead of storing a second outcome value
- removes automatic runtime contract-validation queries for trusted model guards; focused tests own disjointness and completeness obligations
- removes per-family fallback overrides
- keeps enabled-set identity based only on model IDs rather than implementation-specific version fields

No obsolete contract or configuration layer from that list remains in the #383 code path.

### Unresolved `Array.shift` elements

- keep a definitely-array receiver applicable when its element sort is unresolved
- copy the Boolean, number, and address symbolic-memory regions that form the fake-value representation
- read unresolved elements from the current symbolic memory rather than allocation history, including aliased symbolic-index writes
- normalize conditional fake-wrapper/raw-reference values through the existing lazy fake-value machinery without nesting wrappers
- materialize the removed element through the live `TsStepScope`, preserving the exactly-one-type constraint and updated solver models in every successor
- shift already materialized fake wrappers for concrete unknown arrays and preserve reference aliases
- keep allocation history immutable during `shift`; current-state resolution observes the final symbolic memory
- cover mixed `any[]`, empty arrays, all three symbolic regions, aliases, repeated shifts, and multi-element reference arrays

## Scope boundary

This PR completes the common contract and intrinsic integration only. It does not claim source-model execution as delivered.

[...]

Issue #360, opened 2026-08-22
[Epic][TS Calls] Introduce explicit fallback policies and partial semantic models


CaelmBleidd:
## Goal

Make unknown-call behavior explicit and evaluate its effect on symbolic exploration and concrete reproducibility.

## Scope

- Use the current USVM and native JacoDB TypeScript frontend.
- Reuse one typed unknown-call dispatch boundary.
- Select an immutable set of semantic models independently from STOP_PATH or FRESH_SYMBOLIC_RETURN fallback.
- Support guarded normal/exceptional successors, state changes and aliases, with an explicit residual domain routed to fallback.
- Execute source models through the normal EtsIR interpreter; use intrinsics for engine primitives unavailable or inefficient in ordinary TypeScript.
- Observe decisions through TsInterpreterObserver. Persistent path history and stopped-state snapshots are not production requirements.
- Select a small justified model set using a real-project census and existing-model audit, then compare policy/model configurations with real TypeScript replay.

## First-priority gate

**#383 is P0 and blocks completion of #365/#366 and further integration using their old contract.** Reuse the simplification already in PR #377/#380; do not create a duplicate implementation.

A model needs an ID, a declarative target and execution. Do not reintroduce separate backend/registration/descriptor layers, redundant implementation-kind/precision values, manually maintained model versions, or a generic plugin framework.

## Work

- Completed foundation: #361, #362, #363, #364.
- Priority simplification gate: #383.
- Model integration: #365, #366.
- Independent census and audit: #367, #368. Their preparatory work may continue while the gate is resolved.
- Final policy/model experiment: #385. Add bounded family implementation prerequisites only when the census justifies them.
- Optional research-only frontier capture: #373, only when a concrete consumer needs more than existing decision events.

## Definition of Done

- All relevant unknown calls use one documented boundary or have an explicit exclusion.
- Fallback semantics are consistent; opaque returns do not claim to model arbitrary side effects or exceptions.
- Built-in model applicability depends on sufficient target/receiver evidence, not a method name or one possible type.
- Unsupported domains retain their configured fallback; trusted models have focused guard, alias and exception tests.
- One useful intrinsic and one ordinary TypeScript model demonstrate the two implementation mechanisms.
- Decision observation does not change exploration or require generic core-state changes.
- A frozen model set and development/held-out evaluation in #385 measure real replay-confirmed targets on comparable budgets.
- Mandatory language semantics and correctness fixes remain identical between experimental profiles.
[...]

PR #380, opened 2026-08-28
[TS Calls] Execute semantic models written in TypeScript

### LINKED ISSUE #366: [TS Calls] Execute semantic models written in TypeScript
Part of #360. Depends on #365 and **P0 gate #383**.

## Goal

Run semantic models written as ordinary TypeScript through the normal EtsIR interpreter.

## Scope

- Implement a source-model object using the same ID/target/execution contract as #365. Do not add a separate backend registry, registration wrappers or duplicated implementation-kind tags.
- Load the selected model sources through the native JacoDB frontend and add their EtsIR methods to the scene.
- Bind the original receiver and arguments to the model entry point; use normal call execution for return values, exceptions, heap changes and aliases.
- Check the supported domain before entry. Unsupported states retain the configured residual fallback.
- Route nested unknown calls through the same dispatcher and prevent recursive model redirection.
- Implement Array.pop as source code using indexing and length, not by calling pop again. Correct array-length writes must use the engine's actual array-length representation.
- Keep the initial supported receiver/element domain explicit; unsupported/unknown/fake representations fall back. A single possible type returned from a multi-valued stream is not proof.
- Reuse existing frontend loading, scene validation and signature resolution. Cache/loading work should be driven by actual reuse, not a new generic artifact subsystem.
- Record source/artifact content identity where externally loaded models can differ independently of the tool revision. Do not introduce manually managed model versions or compatibility negotiation.

## Definition of Done

- One pure source model and the stateful Array.pop case run end to end.
- Focused tests compare supported source-model behavior with original TypeScript for normal results, exceptions, mutation, aliases and residual inputs.
- Model redirection cannot recursively re-enter the same unsupported call forever.
- Array length, return values and input reconstruction remain consistent.
- #383's simplified configuration and documentation are shared with PR #377.
- Existing PR #380 is adapted and verified; no duplicate implementation is required.

A model backend platform, complete ECMAScript array semantics and unrelated core redesign are outside this issue. Extend domains only with a demonstrated case and tests.

CaelmBleidd:
## Summary

Execute semantic models written in ordinary TypeScript through the existing EtsIR interpreter. `Array.pop` uses array indexing and `length`, preserving numeric, reference, and unresolved elements, mutations, and aliases.

- Load and validate static model entry points with the native JacoDB TypeScript frontend.
- Keep an immutable EtsIR JSON snapshot and materialize a separate object graph for each machine. Models sharing a source file share one materialized file within that machine.
- Bind receiver and arguments through ordinary call frames; preserve returns, exceptions, nested dispatch, and recursion prevention.
- Preserve domain guards, residual fallback, automatic built-in discovery, selected IDs, and signature-collision checks. Model declarations cannot bypass guards through application method lookup.
- Write array length through the normalized storage type. Read numeric fake-value payloads from current memory with their runtime-kind guard, so numeric `any` values and symbolic `any[]` pop results can shrink arrays. Negative zero remains accepted.
- Apply length constraints through `scope.assert`, which updates both path constraints and cached models and stops infeasible paths.

The branch is based on `main` at `a442b2a9`, including #377.

## Supported domain

The built-in `pop` model accepts zero-argument calls on proven one-dimensional arrays, including `any[]`, `unknown[]`, and normalized widened/wrapped aliases. Unsupported inputs use the configured fallback.

Length assignments support integral numeric values from zero through the current length, within the configured size limit. Growth and coercion of nonnumeric lengths remain unsupported; growth and sparse slots are tracked in #388.

## Validation

For head `8188fecd`:

- 54 focused artifact, pop, shift, and EtsIR execution tests passed.
- 13 catalog tests passed in a separate process. Local JDK 23 required MockK self-attach flags.
- The new regression checks cached-model validity after assigning an unconstrained `any` value to array length, with soft-constraint repair disabled. Temporarily replacing both assertions with direct `pathConstraints +=` made that regression fail; the committed implementation retains `scope.assert`.
- `:usvm-ts:detektMain` and `:usvm-ts:detektTest` passed with 0 smells; `git diff --check` passed.
- Full [CI for this head](https://github.com/UnitTestBot/usvm/actions/runs/35448905295) passed: core, JVM, Python, TypeScript, TS PBT, and lint.

Closes #366

Part of #360


usvm-ts/src/main/kotlin/org/usvm/machine/call/TsEtsIrUnknownCallModel.kt
```diff
@@ -0,0 +1,258 @@
...
+import java.nio.file.Path
+import java.security.MessageDigest
+import java.util.IdentityHashMap
+import kotlin.io.path.createTempDirectory
+import kotlin.io.path.deleteIfExists
+import kotlin.io.path.outputStream
+import kotlin.io.path.readBytes
+
+private const val BYTE_MASK = 0xff
+private val sha256Regex = Regex("[0-9a-f]{64}")
```
CaelmBleidd: Ну вот что это за говно? Я не понимаю, зачем ты в каждый реквест тащишь эту залупу? Убери все фингерпринты, ша256, прочую хуету нахуй

usvm-ts/src/main/kotlin/org/usvm/machine/call/TsEtsIrUnknownCallModel.kt
```diff
@@ -0,0 +1,258 @@
...
+        if (inputs.size != artifact.entryPoint.parameters.size) {
+            return null
+        }
+
+        val guard = domainGuard.evaluate(
+            state = state,
+            call = call,
+            inputs = inputs,
+        )
+        if (guard == state.ctx.falseExpr) {
```
CaelmBleidd: isFalse?

usvm-ts/src/main/kotlin/org/usvm/machine/expr/WriteField.kt
```diff
@@ -48,10 +53,81 @@ internal fun TsExprResolver.handleAssignToInstanceField(
...
+    arrayType: EtsArrayType,
+    value: UExpr<*>,
+    maxArraySize: Int,
+): Unit? = with(this) {
+    val (fpLength, numericTypeGuard) = scope.calcOnState {
+        with(ctx) {
+            extractValue(value, fp64Sort, ::getIntermediateFpLValue)
+        }
+    }
+    val numericTypeIsPossible = scope.assert(numericTypeGuard)
```
CaelmBleidd: Это правильное использование ассерта? Может быть такое, что тут ты заасертишь и модель станет false, что приведет к тому, что позже свалятся проверки. Из-за того, что модель стала не консистентной. Добавлять надо через pathConstraints, тогда все будет согласованно

usvm-ts/src/main/kotlin/org/usvm/machine/expr/WriteField.kt
```diff
@@ -48,10 +53,81 @@ internal fun TsExprResolver.handleAssignToInstanceField(
...
+    val lengthIsNonNegative = mkBvSignedGreaterOrEqualExpr(length, mkBv(0))
+    val lengthIsWithinLimit = mkBvSignedLessOrEqualExpr(length, mkBv(maxArraySize))
+    val lengthIsNotGrowing = mkBvSignedLessOrEqualExpr(length, currentLength)
+    val validLength = mkAnd(
+        lengthIsIntegral,
+        lengthIsNonNegative,
+        lengthIsWithinLimit,
+        lengthIsNotGrowing,
+    )
+    scope.assert(validLength) ?: run {
```
CaelmBleidd: То же сомнение про assert

Issue #367, opened 2026-08-22
[TS Calls] Build a real-project call census and model roadmap


CaelmBleidd:
Part of #360. Uses #364; contributes to #385.

## Goal

Measure unknown calls reached on real TypeScript projects and select a small evidence-based model roadmap.

## Scope

- Freeze development-project revisions, licenses, selection criteria, analyzed methods, input domains, frontend/tool versions, seeds and budgets.
- Reuse the existing native frontend, runner and observer. Aggregate decision events outside the engine; #373 snapshots are not required for a basic census.
- Define the observation profile explicitly. A stop-on-first-unknown run cannot reveal later calls on that path. Use a fixed opaque-continuation census and/or a separately reported stop profile to distinguish encountered calls and stopping points.
- Opaque exploration can include infeasible concrete paths and omit callee effects. Report a census of observed events under the profile, not an exhaustive list of all calls or all runtime blockers in a project.
- Normalize API families/methods without conflating same-named user methods with proven built-ins.
- Count projects, distinct call sites, distinct containing functions and dynamic events with explicit denominators. Rank primarily by project prevalence and distinct sites; report loop event counts separately.
- Classify observations as model candidates, frontend/engine defects, external/user behavior or already supported semantics.
- Use #368's inventory to avoid reimplementing existing models. Record candidate domains, residual cases and the rationale for source code versus a genuine engine intrinsic.
- Produce a small P0/P1/P2 list. Select a finite experimental subset before final evaluation; held-out projects never select or tune it.

## Definition of Done

- Machine-readable observations preserve profile, limits, project/source examples and raw counts.
- Source-site identity and aggregation are deterministic; repeated budget-limited runs report observed variability instead of promising identical explored events.
- Stop-profile censoring and opaque-profile approximation are explicit.
- A reviewed model subset is ready for #385; missing implementation work is split into bounded family issues with evaluation dependencies.
- No catalog expansion is implemented in this issue.

The integration examples in #365/#366 are allowed before this roadmap; they validate the mechanism and must not be retroactively presented as census-selected models. Corpus selection and aggregation work can proceed independently of #383.

PR #390, opened 2026-09-20
[TS Calls] Evaluate fallback policies and models with concrete replay

### LINKED ISSUE #385: [TS Calls] Evaluate fallback policies and models with concrete replay
Part of #360.

## Goal

Measure how stopping unknown calls, returning opaque values, and applying a small justified set of semantic models affect target discovery, concrete reproducibility and analysis cost.

## Prerequisites

Depends on #383, #365, #366, #367 and #368.

Use the simplified #365/#366 implementation and the development corpus/inventory from #367/#368. Admit and validate the small experimental model set before freezing evaluation inputs. If a selected family still needs implementation, create a bounded family issue and make it a prerequisite.

## Experiment

- Define three questions: how often unknown calls obstruct analysis; what opaque continuation changes; what validated models add to opaque continuation and stopping.
- Use one USVM/JacoDB revision and one native TypeScript frontend, solver configuration, search policy, input construction and target set for every profile.
- Evaluate the four combinations of an empty/frozen model set and STOP_PATH/FRESH_SYMBOLIC_RETURN. These are experiment configurations, not four engine APIs. Keep mandatory language semantics and correctness fixes identical in all profiles.
- Separate development projects used for model selection from held-out real projects. Freeze revisions, selected methods, target denominator, model IDs/content identity, seeds, phase/total budgets and environment before held-out runs.
- Reuse a concrete runner for the original TypeScript. A target counts as confirmed only when replay reaches that target; observing only a matching return value or exception is insufficient. Reuse existing replay components where suitable without requiring the PBT orchestration, hints or shrinking pipeline.
- Preserve solver-reached targets, extracted inputs, replay-confirmed targets, rejected/unrepresentable candidates, unknown-call decisions, timeouts and tool failures separately.
- Primary measures: unique replay-confirmed targets and their common denominator, confirmation rate, and total runtime. Also report solver-reached targets and unknown-call frequencies by project and distinct call site; solver counters are optional if already available.
- Use repeated seeded runs and a fixed or rotated execution order. Report variability and timeout censoring; equal seeds do not guarantee identical wall-clock-limited exploration.
- Preserve raw per-run results and provide a script producing the tables. Negative results and unsupported cases remain visible.

## Definition of Done

- One command reproduces the experiment from the frozen manifest.
- Minimal fixtures demonstrate lost continuation, a spurious opaque witness and a replay-confirmed modeled witness.
- Development and held-out results distinguish the marginal effect of model selection and residual fallback.
- All reported confirmed coverage uses the same target denominator and real TypeScript replay.
- Conclusions answer the three questions even if modeling gives no gain.
- Model validation, corpus selection and threats to validity are documented.
- Deliver the experiment and report in a dedicated PR. Do not import historical result tables as current evidence.

PBT generation, runtime hints, hybrid scheduling, LLM/MCP and a general model-catalog framework are outside this issue. Stopped-frontier snapshots from #373 are optional and require a demonstrated measurement need.

CaelmBleidd:
The four unknown-call configurations need comparable target measurements and concrete replay against the original TypeScript. This PR adds the frozen experiment runner, preserves symbolic and replay outcomes separately, and stores lossless scalar witnesses so individual results can be reproduced.

It also aligns symbolic target observation with original-source replay. For statements such as `throw new Error(...)`, EtsIR evaluates nested instructions before the exact-span `EtsThrowStmt`; the engine now observes a uniquely proven CFG entry into the complete source-statement lowering region. Ambiguous or insufficient origin metadata is reported as unsupported.

The symbolic engine distinguishes path exhaustion from a configured stop. This keeps timeout classification exact for the experiment without changing the shared `usvm-core` stop-strategy machinery; the experiment enables timeout as its only stopping condition.

The experiment supports the empty/frozen model sets crossed with STOP_PATH/FRESH_SYMBOLIC_RETURN, validates exact project/tool/frontend/source identities, rotates profiles, preserves historical decoding compatibility, and provides raw-result aggregation plus single-witness replay.

Validation:

- source-statement entry tests: 3/3;
- Calls experiment tests: 7/7;
- machine completion-kind regression: 1/1;
- post-#367 census interaction tests: 10/10;
- Detekt for `usvm-ts` and `usvm-ts-pbt`, main and test: zero findings;
- clean `:usvm-ts-pbt:installDist` with exact embedded head;
- installed one-target/four-profile replay: all profiles reached `fibonacciIterative#negative-input` and replayed as confirmed.

Closes #385.


usvm-ts-calls/src/main/kotlin/org/usvm/ts/calls/CurrentTsCallsSymbolicEngine.kt
```diff
@@ -0,0 +1,371 @@
...
+        }
+
+        val modelSelection = if (request.profile.usesFrozenModels) {
+            TsUnknownCallModelSelection.Only(request.frozenModelIds)
+        } else {
+            TsUnknownCallModelSelection.Only(emptySet())
+        }
+        val machineOptions = UMachineOptions(
+            pathSelectionStrategies = listOf(PathSelectionStrategy.BFS),
+            stateCollectionStrategy = StateCollectionStrategy.REACHED_TARGET,
```
CaelmBleidd: Мы же обсуждали использовать не бфс?

